PR #1, opened 2017-08-25
Possible DSL ideas

kdawgwilk:
Came across this today and decided to experiment a little and here is a DSL I came up with.  Not sure what path you want to take but figured I might as well put it up for some feedback or to spark some ideas


orta:
You came across it on the same day I built it, nice timing. This is great 👍 

I'm into it.

PR #3, opened 2017-08-25
Update README for Xcode 9 beta

jpmartha:
For now, developers use Xcode 9 beta. So, I think it's better to add a description for using in Xcode beta.

Feel free to correct my English XD


orta:
👍

PR #14, opened 2017-08-28
Clean up written files on early exit

SD10:
I actually managed to fail the build and found the files weren't being clean up :joy:
I didn't abstract it into a function because it affects the flow/readability of the code.


orta:
Fine with me 👍

PR #15, opened 2017-08-29
Add GitHubMergeRef, GitHubRepo, GitHubIssue, GitHubCommit, GitHubRevi…

SD10:
### Summary of Pull Request:
I ported more of the `danger-js` DSL to `danger-swift`. 
I'm trying to keep the APIs similar but also suitable for Swift.

**Added**:
- GitHubMergeRef
- GitHubRepo
- GitHubIssue
- GitHubIssueLabel
- GitHubCommit
- GitHubReview
- GitCommit
- GitCommitAuthor

**Renamed**:
- Results -> DangerResults

The only thing that's awkward right now is `GitCommit` doesn't have a `sha` or `parents` value when listed under the array of `GitHubCommit` objects. Therefore, both of these properties are optional and that's not ideal in the long run IMO.


SD10:
It would also be useful if you had any more danger-js JSON from a PR with labels and reviews.

orta:
You can use the url https://peril-staging.herokuapp.com/api/v1/pr/dsl?owner=artsy&repo=eidolon&number=609 to generate any PR DSL you'd like :

Sources/Danger/GitHubDSL.swift
```diff
@@ -0,0 +1,329 @@
...
+// MARK: - GitHubIssue
+
+public struct GitHubIssue: Decodable {
+
+    // MARK: - Properties
+
+    /// The labels associated with this issue.
+    public let labels: [GitHubIssueLabel]
+
+}
```
orta: Because of JS' nature of interfaces I can get away with just including the labels on this struct, but realistically it might be worth typing the entire thing - I doubt its going to change in the future

orta: Here's what it looks like: https://github.com/danger/danger-js/blob/master/source/platforms/_tests/fixtures/github_issue.json

SD10: I was curious about that lol. I thought it could be because they could access similar data elsewhere. Not a problem -- I will make these changes 👍 Swift 4 JSON decoding is super easy.

orta:
The CI break is my fault - https://github.com/danger/danger-js/issues/357

SD10:
Thank you for fixing that 💯

PR #16, opened 2017-08-31
Use the alpha of danger

orta:
Makes this project use the alpha, not the latest


orta:
Should fix the CI issue in #15

Issue #20, opened 2017-09-04
Make CI run on Linux


orta:
See https://swiftenv.fuller.li/en/latest/installation.html

ditansu:
@orta is it possible to run a danger-swift on Linux?

orta:
Yep, _our_ CI for this repo is running it on linux. 

There isn't really a distribution method right now though, so you will need to either check out using [linux brew](https://twitter.com/Linuxbrew) or do a manual clone + `make install` then it should be availble

orta:
This is now happening

PR #21, opened 2017-09-04
Refactor danger file path resolver

SD10:
Cleaning up that TODO with a more *"elegant"* solution 😅


orta:
👏

orta:
CI, 20m? Dang - needs #20

orta:
That's fine without an entry

PR #22, opened 2017-09-04
Run CI on Linux

SD10:
I'm not sure if its as easy as this:
https://swiftenv.fuller.li/en/latest/integrations/travis-ci.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,4 +12,6 @@ install:
 
 script:
   - swift test
-  - danger process
+  # - swiftc --driver-mode=swift -L .build/x86_64-unknown-linux/debug -I .build/x86_64-unknown-linux/debug -lDanger Dangerfile.swift fixtures/eidolon_609.json fixtures/response_data.json
+  # - more fixtures/response_data.json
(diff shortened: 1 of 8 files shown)


orta:
Him, I have no idea what that's about

orta:
screen shot 2017-09-04 at 14 51 03
I also get weird errors in https://github.com/JohnSundell/Marathon/pull/127

SD10:
Yeah, I have no idea what these errors are from. I'll keep looking into this

SD10:
@orta Our `Swift.package` requires a Swift tools version of `4.0` but we have Swift 3 declared under `swiftLanaguageVersions`. Is this accurate?

orta:
I guess it depends on your perspective, Swift 4 has both Swift 3 and 4 inside the compiler. TBH though, 4 is a better call there

orta:
Maybe try stealing  this: https://github.com/vapor/vapor/blob/master/circle.yml

SD10:
Thanks for picking up my slack 😅 ❤️

orta:
Heyyyy, that's it working 👍

PR #27, opened 2017-10-16
Makefile

eneko:
This PR adds a `Makefile` to build, install and uninstall `danger-swift`. It will not clone or pull sources from GitHub. This `Makefile` will be used for both manual installation and installation via Homebrew.

## Manual Installation
```
$ git clone https://github.com/danger/danger-swift.git
$ cd danger-swift
$ make install
```

## Uninstall
```
$ make uninstall
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 ## Build generated
 build/
 DerivedData/
+/*.tar.gz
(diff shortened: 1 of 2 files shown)


orta:
👍

PR #28, opened 2017-10-16
Update README.md

eneko:
- Remove Marathon blocker.
- Add instructions for installation via Homebrew tap
- Add instructions for manual installation with `make`.


SD10:
Thank you @eneko ❤️

PR #29, opened 2017-10-16
Handle new lib paths from installing via homebrew

orta:
After looking at https://github.com/Moya/Moya/pull/1377 I got it working, just needs the paths hooking up.

@eneko's changes installed danger the tool correctly, but doesn't handle the other part - deterministically finding `libDanger.dylib` - this PR _should do that_.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,18 @@
 TOOL_NAME = danger-swift
-VERSION = 0.1.0
+LIB_NAME = libDanger.dylib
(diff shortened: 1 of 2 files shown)


orta:
This doesn't quite work, I don't get how to install a library in homebrew

eneko:
Will have to take a look in more detail. Is the library part of the swift package?

eneko:
👏

PR #31, opened 2017-10-18
WIP on generating an xcodeproj

orta:
This sets up the infra to look at editing a dangerfile with auto completion. The idea being that `danger edit` would create an xcodeproject in a temp dir, that references the `Dangerfile.swift` and sets the target paths in the same way as run does.

## TODO

- [ ] Generate an Xcodeproj from scratch
- [ ] Set up the paths same as SPM to allow hooking up to libDanger from different folder


Sources/Runner/Commands/edit.swift
```diff
@@ -0,0 +1,21 @@
...
+
+    // This is a dupe of run
+    let supportedPaths = ["Dangerfile.swift", "danger/Dangerfile.swift", "Danger/Dangerfile.swift"]
+    let resolvedDangerfile = supportedPaths.first { fileManager.fileExists(atPath: $0) }
+
+    let myGroupFile = PBXFileReference(reference: "xxx", sourceTree: .group, path: resolvedDangerfile)
+    myGroup.children.append(myGroupFile.reference)
+
+    project.pbxproj.append(myGroupFile)
+    try project.write(path: "myproject.xcodeproj")
```
orta: Took this from https://github.com/xcodeswift/xcproj/pull/116 - will need to do a bunch of work to generate a project from scratch though

Sources/Runner/main.swift
```diff
@@ -1,134 +1,8 @@
...
-// Check and find where we can link to libDanger from
-let libDanger = "libDanger.dylib"
-let libPaths = libraryFolders + depManagerDangerLibPaths
-
-
-func isTheDangerLibPath(path: String) -> Bool {
-    return fileManager.fileExists(atPath: path + "/libDanger.dylib")  || // OSX
-           fileManager.fileExists(atPath: path + "/libDanger.so")        // Linux
+let cliLength = CommandLine.arguments.count
+if cliLength > 0 && "edit" == CommandLine.arguments[0] {
```
eneko: I thought `CommandLine` was deprecated. Should use `ProcessInfo.processInfo.arguments` instead. Also, consider `arguments[0]` is the executable.

pepibumur:
Have you tried using [XcodeGen](https://github.com/yonaskolb/XcodeGen) instead? XcodeGen provides you with the project specification model that you can pass to the generator and get the project generated for you.

orta:
Yeah, that's probably a better idea IMO - I think we can demand that to use the feature, mint must be installed - then can be mint's job to download and run XcodeGen and all danger-swift has to do is create the right project JSON file. Moving this into an issue.

orta:
Moved to #35

PR #36, opened 2017-11-25
Adds an edit command

orta:
This uses MarathonCore to create an edit command - it's rough, but it works 👍 

Depends on https://github.com/JohnSundell/Marathon/pull/148


orta:
screen shot 2017-11-25 at 1 31 50 pm

PR #37, opened 2017-11-26
Add support for plugins

orta:
Uses Marathon to download and setup any required dependencies, which is basically the infra for plugins. It will then hook up the dependencies at runtime at the same way that libDanger is hooked up.

Requires on https://github.com/JohnSundell/Marathon/pull/148


orta:
OK 👍

PR #42, opened 2017-12-04
Generate Dangerfile for edit command when it DNE

### LINKED ISSUE #38: Danger Edit should create a Dangerfile for you, if you don't have one
This can simplify the setup for first runs.

SD10:
This could possibly resolve #38.

When generating the Dangerfile I added the boiler plate template:
```Swift
import Danger

let danger = Danger()
```

Also, question about the new DSL. The test JSON in [fixtures/eidolon_609.json](https://github.com/danger/danger-swift/blob/master/fixtures/eidolon_609.json) has `danger` double nested. 

Does the accurately DSL represent this? I had trouble evaluating with the commands on README 😕

### DIFF
--- a/Sources/Runner/Commands/Edit.swift
+++ b/Sources/Runner/Commands/Edit.swift
@@ -5,13 +5,20 @@ import Files
 import MarathonCore
 
 func editDanger() throws -> Void {
-    
-    // Exit if a dangerfile was not found at any supported path
-    guard let dangerfilePath = Runtime.getDangerfile() else {
-        print("Could not find a Dangerfile")
-        print("Please use a supported path: \(Runtime.supportedPaths)")
-        exit(1)
+
+    let createDangerfile = { () -> String in
+        do {
+            let template = "import Danger \n let danger = Danger()"
+            let data = template.data(using: .utf8)!
+            return try FileSystem().createFile(at: "Dangerfile.swift", contents: data).path
+        } catch {
+            print("Could not find or generate a Dangerfile")
+            exit(1)
+        }
     }
+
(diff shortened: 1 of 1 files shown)


orta:
Feature looks good: 🎫 

re:double danger - Yeah, I noticed this discrepancy when I was building out something recently. It doesn't make sense, I'm definitely sure that's what the Danger DSL looks like

screen shot 2017-12-04 at 7 49 33 am

As the first one represents data for danger, and the second one represents the danger export. Is it possible that when being passed for the compiler to the runner sub-process we remove one of those dangers?

Sources/Runner/Commands/Edit.swift
```diff
@@ -5,14 +5,21 @@ import Files
...
-        print("Please use a supported path: \(Runtime.supportedPaths)")
-        exit(1)
+
+    let createDangerfile = { () -> String in
+        do {
+            let template = "import Danger \n let danger = Danger()"
+            let data = template.data(using: .utf8)!
+            return try FileSystem().createFile(at: "Dangerfile.swift", contents: data).path
+        } catch {
+            print("Could not find or generate a Dangerfile")
```
BasThomas: Maybe print the error while you're at it?

SD10: I thought about this but wasn't sure if it was right to propagate error messages coming from the `Files` module. I didn't want to clog up stdout

SD10:
@orta What if we trim 13 leading bytes and 2 trailing bytes off the incoming DSL JSON?  It seems hacky but I think converting the JSON to a `String` to process it is overkill

PR #43, opened 2017-12-04
Strip top level danger from DSL JSON

SD10:
As discussed in #42 we could trim the bytes off the DSL JSON data until the top level danger object is removed. 

The code is pretty nasty but it basically:
- deletes all bytes up until the second `{` from the start
- removes all bytes up to and including the first `}` from the end

It's 13 leading bytes and 2 trailing bytes... but this method should account for any fluctuation

### DIFF
--- a/Sources/Danger/Danger.swift
+++ b/Sources/Danger/Danger.swift
@@ -28,13 +28,14 @@ private final class DangerRunner {
             exit(1)
         }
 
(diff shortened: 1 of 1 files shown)


SD10:
Ok looks like I misunderstood the issue 😅 😅

orta:
Yeah, think it is that the STDIN that gets passed into the runner side might have that first danger stripped, not sure, don't have time to look in

PR #47, opened 2017-12-12
Add Milestone and improve more

d-date:
### Motivation and Context
We sometimes need to validate pr whether milestone is set, so I've prepared milestone model.

### Description

Other changes as below

- [x] add milestone model
- [x] change iso8601 date string type to `Date` type using `.iso8601` date decoding strategy (all date string get from GitHub API are formatted by `iso8601` )
- [x] resolved DecodingError of `fixtures/eidolon_609.json`.
- [x] change iso6801 on comments, which means [rubbers and plastic houses](https://www.iso.org/standard/13302.html) to iso8601.

And I've added milestone structure on `fixtures/eidolon_609.json` to check decoding json correctly.


Sources/Danger/GitHubDSL.swift
```diff
@@ -91,17 +92,17 @@ public struct GitHubPR: Decodable {
...
     public let assignees: [GitHubUser]
 
-    /// The ISO6801 date string for when the pull request was created.
-    public let createdAt: String
+    /// The ISO8601 date string for when the pull request was created.
+    public let createdAt: Date
 
-    /// The ISO6801 date string for when the pull request was updated.
-    public let updatedAt: String
+    /// The ISO8601 date string for when the pull request was updated.
```
SD10: The documentation for all these still reference `String` after you've changed the type to `Date`

Sources/Danger/GitHubDSL.swift
```diff
@@ -91,17 +92,17 @@ public struct GitHubPR: Decodable {
     /// The users who are assigned to the pull request.
     public let assignees: [GitHubUser]
 
-    /// The ISO6801 date string for when the pull request was created.
-    public let createdAt: String
+    /// The ISO8601 date string for when the pull request was created.
```
SD10: I got all the comments from `danger-js` so these could use a cleaning up there too 😅 
https://github.com/danger/danger-js/blob/9a501500ee81e593e53062dab1d723908454ac81/source/dsl/Commit.ts#L25

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 * Add your own contributions to the next release on the line below this, please include your name too. Please don't set
   a new version if you are the first to make the section for `master`.
+* Add milestone model to issue and pull request. - d-date
```
SD10: Not sure if we want to mention that we broke the api with `String` -> `Date`. The project is still quite young

d-date:
@SD10 Thank you for quick review! I fixed all😄

Sources/Danger/GitHubDSL.swift
SD10: @d-date `The ISO8601 date string for when` -> `The date when`
Very nitty, I know

orta:
lol 🦆

orta:
Ah, yeah, this is something I hadn't really thought of - in DangerJS we only provide interfaces for the objects, but it's fine to access keys in the DSL without specifying them into the DSL. Looks like there'll be more easy low-hanging fruit like this 👍

PR #50, opened 2017-12-27
Logger struct, --verbose and --silent

rockbruno:
Hey there, I saw the issue asking for a `verbose` argument, so I made a simple logger struct to support it, together with two arguments: `--verbose` and `--silent`, slightly based on `Cork`

I was not sure how to inject `Logger` into `DangerRunner` since it runs separately, so I forwarded all arguments to it and instanced a new Logger there. I'm not sure if this is the best way to approach it, so criticize all you want!


orta:
@rockbruno  - yeah this seems 👍 to me TBH, it's definitely better than before

Issue #52, opened 2018-01-30
Bad JSON input to danger-swift (possibly due to Github API change?)


hirad:
Starting today, I noticed that Danger is failing on all our PRs. It turned out to be an issue with parsing the JSON input to `danger-swift`:

```
Failed to parse JSON:
keyNotFound(Danger.DangerDSL.(CodingKeys in _125A41DBF2E88FF661E0539497E789E1).git, Swift.DecodingError.Context(codingPath: [Danger.DSL.(CodingKeys in _125A41DBF2E88FF661E0539497E789E1).danger], debugDescription: "No value associated with key git (\"git\").", underlyingError: nil))
```

When I ran Danger locally (using `danger pr https://github.com/path/to/pr --json`), I saw that the generated JSON had two initial `danger` keys:

```
{
  "danger": {
    "danger": {
      "git": { ... }
    }
  }
}
```

If I edit that file and remove one of the `danger` keys, then we get further but not by much:

```
Failed to parse JSON:
typeMismatch(Swift.Array<Any>, Swift.DecodingError.Context(codingPath: [Danger.DSL.(CodingKeys in _125A41DBF2E88FF661E0539497E789E1).danger, Danger.DangerDSL.(CodingKeys in _125A41DBF2E88FF661E0539497E789E1).github, Danger.GitHub.CodingKeys.requestedReviewers], debugDescription: "Expected to decode Array<Any> but found a dictionary instead.", underlyingError: nil))
```

This last issue seems to be related to [this breaking change](https://developer.github.com/changes/2018-01-24-end-thor-preview/) in the Github API. I'll be happy to open a PR for this to address, but at the moment, I'm not sure what the first problem is about. Some guidance on debugging it further would be appreciated :) 

Edit: the double-`danger` key seems to be an issue with running it locally. I just verified that that doesn't seem to occur on CI. On CI I'm only seeing the `requestedReviewers` issue. I'll try to spend some time on that this week.

orta:
Yeah, this whole double danger thing is still a little bit confusing to me - WRT the API thing, yeah, we'll have to change the types and ship a new release of danger-swift

orta:
This is fixed 👍

PR #53, opened 2018-01-30
Support team review requests

hirad:
My attempt at fixing #52. Seems to work with sample data obtained [here](https://developer.github.com/v3/pulls/review_requests/#list-review-requests).

### DIFF
--- a/Sources/Danger/GitHubDSL.swift
+++ b/Sources/Danger/GitHubDSL.swift
@@ -189,10 +189,10 @@ public struct GitHubTeam: Decodable {
 
     // MARK: - Properties
 
-    // The UUID for the team.
+    /// The UUID for the team.
     public let id: Int
 
-    // The team name
+    /// The team name
     public let name: String
 }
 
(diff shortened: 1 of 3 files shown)


Sources/Danger/GitHubDSL.swift
```diff
@@ -175,6 +175,45 @@ public struct GitHubUser: Decodable {
...
+    // MARK: - CodingKeys
+
+    enum CodingKeys: String, CodingKey {
+        case users
+        case teams
+    }
+
+    // MARK: - Properties
+
+    let users: [GitHubUser]
```
SD10: This needs to be marked public and it would be great if you add comments 🙏

Sources/Danger/GitHubDSL.swift
```diff
@@ -175,6 +175,45 @@ public struct GitHubUser: Decodable {
...
+    let teams: [GitHubTeam]
```
SD10: Ditto

orta:
Awesome

orta:
OK, this should be shipped as 0.3.2

PR #56, opened 2018-03-04
Inline mode

sunshinejr:
(This needs to wait before [danger-js#529](https://github.com/danger/danger-js/pull/529) is merged, though)

Live demo: [Harvey#17](https://github.com/Moya/Harvey/pull/17)

#### Todo:
- [ ] Add documentation
- [ ] Add Changelog entry

### DIFF
--- a/Sources/Danger/Danger.swift
+++ b/Sources/Danger/Danger.swift
@@ -74,6 +74,11 @@ public func warn(_ message: String) {
     DangerRunner.shared.results.warnings.append(Violation(message: message))
 }
 
+/// Adds an inline warning message to the Danger report
+public func warn(message: String, file: String, line: Int) {
+    DangerRunner.shared.results.warnings.append(Violation(message: message, file: file, line: line))
+}
+
 /// Adds a warning message to the Danger report
 ///
 /// - Parameter message: A markdown-ish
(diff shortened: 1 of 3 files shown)


Sources/Danger/Danger.swift
```diff
@@ -74,25 +74,45 @@ public func warn(_ message: String) {
...
+
 /// Adds a warning message to the Danger report
 ///
 /// - Parameter message: A markdown-ish
 public func fail(_ message: String) {
     DangerRunner.shared.results.fails.append(Violation(message: message))
 }
 
+/// Adds an inline fail message to the Danger report
+public func fail(message: String, file: String, line: Int) {
```
SD10: Any reason you chose to use function overloading instead of just making file & string parameters optional with defaults?

sunshinejr: Oh, yeah. I just don't think we should let users add the file without a line. Although it's possible to do (in danger-js, currently, because I couldn't find a better way without over-engineering), it doesn't do much (will just end up in the main comment anyways).

SD10: Great point 👍 Was just curious and wanted to understand your reasoning

sunshinejr: Totally, didn't think about that before, good question 👍

orta: Yeah, I like this 👍

orta:
I'll handle the rest

PR #57, opened 2018-03-09
Reordering how Runner args are routed to Danger

rockbruno:
It seems like Marathon's number of arguments cannot be predicted, which caused https://github.com/danger/danger-swift/issues/54 😫 

The reverting PR https://github.com/danger/danger-swift/pull/55 fixes the issue, but also makes `--verbose` and his friends stop working.

It seems that injecting the arguments right after the Dangerfile will hopefully fix this once and for all. Since we only check the last arguments, hopefully injecting them in the middle will not cause issues. We can also drop logging entirely from the second part if it makes more sense, so we won't have to do this weird arg routing thing.

@sunshinejr If possible, could you help me test this? I think you have a setup where Marathon ends up with different arguments than me, which would be a way to confirm that this won't break things again

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ## Master
 
+* Reordering how Runner args are routed to Danger - rockbruno
+
 ## 0.3.3
 
 * Fixes for the CLI arg order from danger-js - sunshinejr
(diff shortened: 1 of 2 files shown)


orta:
Waiting on @sunshinejr  - but yeah, this seems 👍

sunshinejr:
@rockbruno oh, now it makes sense! I've tested in on [Harvey](https://circleci.com/gh/Moya/Harvey/177) (I had problems on there), but it seems like it's all fine!

@orta should be good to go 👍

orta:
Shipped as 0.3.4

PR #61, opened 2018-03-24
Tests for GitHubDSL model objects

yhkaplan:
## Purpose
To make sure Danger-Swift is tested and therefore easy to maintain going forward.

## Changes
1. Wrote test for GitHubUser and one test for GitHubMilestone
2. Moved the DateFormatter used in GitHubDSL to an extension so tests could use the exact same one
3. Used Sourcery to autogenerate equatable extensions to all GitHub and Git models
    * I used an example template from the Sourcery project to generate equatable extensions, adding only an import statement to the template
    * I also added a Sourcery config file to make using Sourcery easier
    * All you have to do to update the models is: `$ sourcery --config .sourcery.yml`
4. Fixed an issue with GitHubMilestone where properties were not marked public

## Plans
1. Adding tests for all the other model types
2. Deleting the autogenerated equatable extensions once Swift 4.1 is released (which generates equatable conformance for free)

## Areas to consider 
1. Equatable
Normally, object with an ID property should be considered equatable with just the same IDs. People making plugins will likely want the models to be equatable, so I propose that at a certain point we make the regular equatable extensions looking only at the id where relevant. Then, in the test target we can use code generation to create convenience methods for objects that conform to a custom protocol named `FullyEquatable` or something like that.

2. How to handle test resource files
SPM does not allow for test resource files (like json files), so, after examining the thread below, I put the JSON in a big multiline string literal inside Swift files. The solutions in the thread below all looked very hacky, and I thought mine was more simple and stable, but please tell me your thoughts. (I think Apple will release support for test resources for SPM at some point because they plan on using server-side Swift; it's just a matter of when.)
https://stackoverflow.com/questions/47177036/use-resources-in-unit-tests-with-swift-package-manager
https://bugs.swift.org/browse/SR-2866

## Related
#9


yhkaplan:
Hmm, not really sure what the CI error is: anyone have a guess?

```
Error:  TypeError: Cannot read property 'concat' of undefined
    at Executor.<anonymous> (/Users/travis/.nvm/versions/node/v6.11.5/lib/node_modules/danger/distribution/runner/Executor.js:236:45)
    at step (/Users/travis/.nvm/versions/node/v6.11.5/lib/node_modules/danger/distribution/runner/Executor.js:32:23)
    at Object.next (/Users/travis/.nvm/versions/node/v6.11.5/lib/node_modules/danger/distribution/runner/Executor.js:13:53)
    at /Users/travis/.nvm/versions/node/v6.11.5/lib/node_modules/danger/distribution/runner/Executor.js:7:71
    at __awaiter (/Users/travis/.nvm/versions/node/v6.11.5/lib/node_modules/danger/distribution/runner/Executor.js:3:12)
    at Executor.handleResultsPostingToPlatform (/Users/travis/.nvm/versions/node/v6.11.5/lib/node_modules/danger/distribution/runner/Executor.js:230:16)
    at Executor.<anonymous> (/Users/travis/.nvm/versions/node/v6.11.5/lib/node_modules/danger/distribution/runner/Executor.js:163:26)
    at step (/Users/travis/.nvm/versions/node/v6.11.5/lib/node_modules/danger/distribution/runner/Executor.js:32:23)
    at Object.next (/Users/travis/.nvm/versions/node/v6.11.5/lib/node_modules/danger/distribution/runner/Executor.js:13:53)
    at /Users/travis/.nvm/versions/node/v6.11.5/lib/node_modules/danger/distribution/runner/Executor.js:7:71
The command "danger process .build/x86_64-apple-macosx10.10/debug/danger-swift" exited with 1.
6.34s$ sudo chmod -R a+rwx /usr/local/
```

Tests/DangerTests/GitHubTests.swift
```diff
@@ -1,14 +1,73 @@
...
+        
+        XCTAssertEqual(testMilestone, correctMilestone)
+    }
+    
+    func test_GitHubMilestone_decodeWithAllParameters() throws {
+        
+    }
 }
 
+public protocol AutoEquatable {}
```
yhkaplan: Still considering where would be the appropriate place for this protocol

Sources/Danger/GitHubDSL.swift
```diff
@@ -483,38 +483,38 @@ public struct GitHubMilestone: Decodable {
     // MARK: - Properties
 
     /// The id number of this milestone
-    let id: Int
+    public let id: Int
```
SD10: Good catch 😱 Can't believe we missed these

yhkaplan: It's easy to mix up 😉

Tests/DangerTests/GitHubTestResources/GitHubUser.swift
```diff
@@ -0,0 +1,6 @@
+
+public let GitHubUserJSON = """
```
SD10: I'm not against this but honestly just running the tests with `swift test` would suffice 🤔

Tests/DangerTests/GitHubTests.swift
```diff
@@ -1,14 +1,73 @@
...
+    func test_GitHubUser_decode() throws {
+        guard let data = GitHubUserJSON.data(using: .utf8) else {
+            XCTFail("Cannot generate data")
+            return
+        }
         
+        let correctUser = GitHubUser(id: 25879490, login: "yhkaplan", userType: .user)
+        let testUser: GitHubUser = try JSONDecoder().decode(GitHubUser.self, from: data)
+        
+        XCTAssertEqual(testUser, correctUser)
```
SD10: Do we really need to check the accuracy of the information? I mean `Decodable` handles the routing of the values to their respective properties. I think danger's main concern is: can we handle the JSON? Thus, I was really only expecting not `nil` testing to assure we handled the values properly.

I also want to add that this makes the tests really brittle if we want to change the underlying JSON.

Tests/DangerTests/GitHubTestResources/GitHubUser.swift
yhkaplan: I think I misinterpreted what you were saying before. Essentially, the test bundle is just a requirement by xcode/xcodebuild, but with `swift test` the json file could be easily found without using a build script or makefile?

SD10: It's fine I likely confused you because I know so little about what I'm talking about 😂 I think the root directory when using `swift test` is the location of the package manifest?

Tests/DangerTests/GitHubTests.swift
yhkaplan: I think that checking the values is useful for three things:
1. Checking that the string value of a CodingKey is not assigned to the wrong case, which is possible in a particularly large CodingKeys enum where different types of ID can exist.
2. Ensuring that optional values that had incorrect CodingKeys are not always set to nil (I plan on writing a test for GitHubMilestone with non-nil values for dueAt, etc)
3. Making sure that properties have public access (as discovered above)

I didn't anticipate having to change the underlying JSON on a particularly frequent basis, only when models are changed, which I thought could reasonably require updates to their test cases. Ultimately though, it depends if you think that the three merits above outweigh extra brittleness for future changes.

Sources/Danger/New Group/DateFormatterExtensions.swift
```diff
@@ -0,0 +1,12 @@
...
+extension DateFormatter {
+    // A default implementation of an iso8601 DateFormatter
+    public static var defaultDateFormatter: DateFormatter {
+        let dateFormatter = DateFormatter()
+        dateFormatter.locale = Locale(identifier: "en_US_POSIX")
+        dateFormatter.timeZone = TimeZone(secondsFromGMT: 0)
+        dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ssZZZZ"
+        return dateFormatter
+    }
+}
```
orta: nice move

Sources/Danger/GitHubDSL.swift
orta: Yeah, Swift's accessor stuff is a pain

Sources/Templates/AutoEquatable.stencil
```diff
@@ -0,0 +1,64 @@
...
+        {% else %}
+        {% for associated in case.associatedValues %}if lhs.{{ associated.externalName }} != rhs.{{ associated.externalName }} { return false }
+        {% endfor %}return true
+        {% endif %}
+        {% endif %}
+    {% endfor %}
+    {% if type.cases.count > 1 %}default: return false{% endif %}
+    }
+}
+{% endfor %}
```
orta: Can you please add docs (the readme or contributing) somewhere on how to update these, I've never used it

Tests/DangerTests/GitHubTests.swift
orta: Yeah,  I can see these arguments, I think that's a pretty good case.

Tests/DangerTests/GitHubTests.swift
orta: `/shrug`

petrpavlik:
@yhkaplan the CI error seems to be generated by the javascript implementation or danger. What I find strange is that danger-swift seems to generate a different `danger-response.json` on macOS and linux, that's why it fails just on macOS. Access level issue? Invalid path? Invalid format of the json? Hard to say, it'd be nice if danger-js could provide a more reasonable error message.

orta:
That's a raised exception, I'm willing to bet that somehow danger-swift is sending a response back that  danger-js doesn't expect. Ideally there should be some verification on the response on danger-ja's side

Tests/DangerTests/GitHubTests.swift
sunshinejr: maybe a new file that would cover all the testing helpers?

yhkaplan: That would make sense! Then I could put that file in the EquatableExtensions folder...

Sources/Templates/AutoEquatable.stencil
yhkaplan: Sure thing! (Sourcery is available on Homebrew, but I kind of wish there was a way to include it in the project directory like with a Gemfile....)

Tests/DangerTests/GitHubTests.swift
SD10: I think 2 and 3 would be covered by testing for nil alone but I can see case 1. Thanks for sharing your thought process and taking care of this one so thoroughly

Tests/DangerTests/GitHubTests.swift
yhkaplan: Fixed in this commit: 7636e5050f7a3b849b4f29f8e3dcdaed7ce464fe

Sources/Templates/AutoEquatable.stencil
yhkaplan: Added in this commit: 66ad8fccaeb3965384f21207fb8da53aa1550e8f

Tests/DangerTests/GitHubTests.swift
yhkaplan: No problem, thanks for the impression —it helped me think through what I was doing more.

Tests/DangerTests/GitHubTestResources/GitHubUser.swift
yhkaplan: I'll look more into this on Monday or Tuesday!

yhkaplan:
I plan on adding tests for more models on Monday or Tuesday. Should I add them as part of this PR or should I make a separate PR? Also, it looks like the CI is still not liking my PR 🤔

orta:
Does it work via `swift test` locally?

yhkaplan:
@orta Running Swift test locally works, with all tests being run, and all passing. Also, looking at the CI output, all the tests appear to execute and pass on both Linux and macOS.

yhkaplan:
GitHubRequestedReviewers' comment states that it represents a value in a PR. However, it's not implemented as a parameter of GitHubPR at this time, so it doesn't appear to need a test at this time. Once it is implemented as part of a PR, GitHubRequestedReviewers is just a simple struct for User and Team, so I think I'll just test to see that it's not nil.

```swift
/// Represents the payload for a PR's requested reviewers value.
public struct GitHubRequestedReviewers: Decodable {

    // MARK: - CodingKeys

    enum CodingKeys: String, CodingKey {
        case users
        case teams
    }

    // MARK: - Properties

    /// The list of users of whom a review has been requested.
    public let users: [GitHubUser]

    /// The list of teams of whom a review has been requested.
    public let teams: [GitHubTeam]
}
```

orta:
I'll take a look on Danger JS's side 👍

yhkaplan:
I added more tests for now, but I can move them to a new PR if that's better. 

One thing I noticed is that GitHubPR couldn't be decoded with two samples of JSON data that I tried. I made several properties optional to fix this in 0c5e2fb4389a0a3aab415862c2b575670176f855 , but I suspect some need new CodingKeys.

orta:
I've done https://github.com/danger/danger-js/pull/548 which should raise if there's anything Danger JS doesn't expect in the JSON passed between danger-js and danger-swift

orta:
Interesting that it's green, do you feel like you changed anything that would have fixed it?

yhkaplan:
@orta No, I don't think any of my recent changes were the reason it was successful this time because they were mostly just adding a couple tests and a lot of test data. I didn't push any configuration changes...

orta:
OK, I'm going to release a new Danger JS, which I'm like 80% sure will break danger-swift. So I'll take this PR, and ship a new release to match the new danger-js one with #56

Issue #72, opened 2018-05-14
`make install` is failing


hiragram:
My environment:

- macOS 10.12.6
- Xcode 9.2
- Swift 4.0.3

```sh
$ make install
swift package clean
swift build --disable-sandbox -c release --static-swift-stdlib
Fetching https://github.com/JohnSundell/ShellOut.git
Fetching https://github.com/JohnSundell/Require.git
Fetching https://github.com/JohnSundell/Marathon.git
Fetching https://github.com/JohnSundell/Wrap.git
Fetching https://github.com/JohnSundell/Files.git
Fetching https://github.com/JohnSundell/Releases.git
Fetching https://github.com/JohnSundell/Unbox.git
error: dependency graph is unresolvable; found these conflicting requirements:

Dependencies:
    https://github.com/JohnSundell/Marathon.git @ 3.0.0..<4.0.0
error: product dependency 'MarathonCore' not found
make: *** [build] Error 1
```

What can I do ? Both of manual install and `brew install` are failing due to the same error.

orta:
Looks like something changed in marathon?

hiragram:
@sunshinejr I haven't tried 9.3 because I am still on Low Sierra. Will consider to upgrade to Higher one. Thank you. I am glad to hear that.

orta:
Maybe you could send a PR checking for that in `make install` and let people know it's 9.3 only?

hiragram:
Nice. I will.

hiragram:
For now, I have updated readme as telling that Swift 4.1. is required.

orta:
Cool, this PR got merged

PR #74, opened 2018-05-22
Add requirements (Swift version)

hiragram:
Related to #72

Latest release of danger-swift only works with Swift 4.1 .


orta:
Good thinking

Issue #76, opened 2018-05-27
Should sourcery be removed from code?


Sega-Zero:
Swift 4.1 already has Auto-equatable feature, so is there still a reason to have sourcery template for a boilerplate code?

orta:
That could work yeah, less is less

yhkaplan:
I’m the one who added Sourcery generated files. I’ll try removing them when I get the chance.

yhkaplan:
Just tried this in a personal project. One limitation seems to be that Swift won’t generate conformance in the test target, separate from the main target, but I don’t foresee any issue with adding equatable conformance directly in the main target.

PR #78, opened 2018-07-07
Remove sourcery

yhkaplan:
## What 
As per [#76](https://github.com/danger/danger-swift/issues/76), this removes Sourcery templates and code in favor of Swift 4.1 automatic Equatable conformance.

## Why
Code generation for Equatable is not necessary anymore as Swift 4.1 features automatic conformance for Equatable. While Sourcery is reliable, using native Swift auto-conformance makes for a simpler toolset when getting started with Danger, and is significantly less likely to require maintenance between major Swift versions. 

## How
* Remove templates, sourcery.yml, AutoEquatable protocol, and conformance declarations
* Add Equatable with Xcode's regex find and replace as below

## Concerns
* Adding Equatable conformance in the Test target did not work, meaning it had to be added directly in the Main target
    * This adds a small amount of clutter to the model objects affected
* Some model objects feature an "ID" or otherwise unique property that is technically the only parameter needed to define equatability
    * These model objects may have this more technically "correct" conformance added, causing tests to miss all the other parameters
    * For now, I believe this doesn't present a major problem, but it may cause tests coverage to drop dramatically if not watched
    * One solution may be to write a Danger rule that's triggered by `func == (rhs: ...)`, reminding the developer that the tests require a workaround


yhkaplan:
CI failing at Danger JS, but I don't understand what "Danger JS did not include fails" means. Perhaps the config file has some kind of error?
```
Error:  Error: Results passed to Danger JS did not include fails.
{}
    at /Users/travis/.nvm/versions/node/v6.12.3/lib/node_modules/danger/distribution/dsl/DangerResults.js:20:19
    at Array.forEach (native)
    at Object.validateResults (/Users/travis/.nvm/versions/node/v6.12.3/lib/node_modules/danger/distribution/dsl/DangerResults.js:16:24)
    at Executor.<anonymous> (/Users/travis/.nvm/versions/node/v6.12.3/lib/node_modules/danger/distribution/runner/Executor.js:162:41)
    at step (/Users/travis/.nvm/versions/node/v6.12.3/lib/node_modules/danger/distribution/runner/Executor.js:32:23)
    at Object.next (/Users/travis/.nvm/versions/node/v6.12.3/lib/node_modules/danger/distribution/runner/Executor.js:13:53)
    at /Users/travis/.nvm/versions/node/v6.12.3/lib/node_modules/danger/distribution/runner/Executor.js:7:71
    at __awaiter (/Users/travis/.nvm/versions/node/v6.12.3/lib/node_modules/danger/distribution/runner/Executor.js:3:12)
    at Executor.handleResults (/Users/travis/.nvm/versions/node/v6.12.3/lib/node_modules/danger/distribution/runner/Executor.js:158:16)
    at Object.<anonymous> (/Users/travis/.nvm/versions/node/v6.12.3/lib/node_modules/danger/distribution/commands/utils/runDangerSubprocess.js:108:47)
The command "DEBUG="*" danger process .build/x86_64-apple-macosx10.10/debug/danger-swift" exited with 1.
```

orta:
This feels like a danger-js bug tbh

yhkaplan:
Seems to be ok on the Linux CI and my local machine

orta:
I shipped an update with danger-js for some improved logging, but I don't think that should block this PR

PR #79, opened 2018-07-17
[BitbucketServer] Make description, commiter and committerTimestamp optional

acecilia:
Following implementation in https://github.com/danger/danger-swift/pull/75 and https://github.com/danger/danger-swift/pull/73


orta:
👍

Issue #80, opened 2018-07-21
Add support for filetype


yhkaplan:
How would you feel about adding a filetype property to modifiedFiles, createdFiles, and deletedFiles? One common flow I have with Danger is to check the file type. However, this is somewhat painful in Swift because the regex support is significantly less convenient than Ruby's.

It could be implemented in a number of ways. For example, the method below.

```swift
public struct Git: Decodable, Equatable {
    //...
    public let modifiedFiles: [File]
    public let createdFiles: [File]
    public let deletedFiles: [File]
}

public typealias File = String

public extension File {
    var fileType: FileType? {
        return FileType(from: self)
    }
}

public enum FileType: String {
    case storyboard
    case m
    case h
    case swift
    case markdown = "md"
    
    public var `extension`: String {
        return rawValue
    }
    
    static internal var allCases: [FileType] {
        return [.storyboard, .m, .h, .swift, .markdown]
    }

    init?(from fileName: String) {
        let allCasesDelimited = FileType.allCases.map { $0.extension }.joined(separator: "|")
        
         // regex in Swift is so ugly...
        guard
            let regex = try? NSRegularExpression(pattern: "\\.(\(allCasesDelimited))$", options: []),
            let match = regex.firstMatch(in: fileName, options: [], range: NSRange(location: 0, length: fileName.count)),
            let range = Range(match.range, in: fileName)
        else {
            return nil
        }
        
        let rawValue = String(fileName[range])
        self.init(rawValue: rawValue)
    }
}
```

yhkaplan:
I'm thinking something like the below. One issue, however, is that objects of type File that used to be String would have to be converted back and forth, which may cause disruption to existing Dangerfiles. 

However, we could make File easier to deal with by adding an extended property called Name, which would be a String version of itself.

```swift
import Danger

let codeWasModified = git.modifiedFiles
    .filter { $0.fileType == .swift || $0.fileType == .m  || $0.fileType == .h }
    .count > 0

let changelog: File = "Changelog.md"
let changelogWasModified == git.modifiedFiles
    .filter { $0 == changelog } // $0 == "some string" probably wouldn't work here anymore
// Alternative to above: .filter { $0.name == "Changelog.md" }
    .count > 0

if codeWasModified && !changelogWasModified {
    warn("Please update the changelog")
}
```

orta:
Thanks - this is in and shipped 👍

PR #81, opened 2018-07-22
Add filetype extensions to modifiedFiles, createdFiles, and deletedFiles

yhkaplan:
As per [#80](https://github.com/danger/danger-swift/issues/80), I propose to change modifiedFiles, createdFiles, and deletedFiles to be of type `File` from String, adding Name and FileType properties. 

This will break some Dangerfile workflow, requiring users to potentially have to access the .name property to get a regular string. In return, users a get type-safe enum property called fileType for .h, .json, .m, .markdown, .mm, .pbxproj, .plist, .storyboard, .swift, .xcscheme, .yaml, .yml. 

In Ruby, detecting the filetype is rather straightforward because of easy-to-use regex support built right in. Swift, on the other hand, would require users to use the clunky NSRegularExpression class, which is bothersome to do inside the Dangerfile. 

### Example of use in Dangerfile
```swift
import Danger

let codeWasModified = git.modifiedFiles
    .filter { $0.fileType == .swift || $0.fileType == .m  || $0.fileType == .h }
    .count > 0

let changelogWasModified == git.modifiedFiles
    .filter { $0.name == "Changelog.md" }
    .count > 0

if codeWasModified && !changelogWasModified {
    warn("Please update the changelog")
}
```

### DIFF
--- a/Sources/Danger/File.swift
+++ b/Sources/Danger/File.swift
@@ -36,7 +36,7 @@ extension FileType {
         return rawValue
     }
 
-    public init?(from file: File) {
+    init?(from file: File) {
         let allCasesDelimited = FileType.allCases.map { $0.extension }.joined(separator: "|")
 
         guard
--- a/Sources/Danger/File.swift
+++ b/Sources/Danger/File.swift
@@ -28,7 +28,7 @@ public enum FileType: String, Equatable {
     }
 }
 
-// MARK: - Public extensions
+// MARK: - FileType extensions
 
 extension FileType {
 
--- a/Sources/Danger/File.swift
+++ b/Sources/Danger/File.swift
@@ -35,16 +35,13 @@ extension FileType {
     public init?(from fileName: String) {
         let allCasesDelimited = FileType.allCases.map { $0.extension }.joined(separator: "|")
 
-        // TODO: move ugly regex stuff to tested extension where it belongs, write tests
         guard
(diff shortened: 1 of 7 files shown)


yhkaplan:
I must have bad luck: `Error:  Error: Results passed to Danger JS did not include fails.`

Sources/Danger/File.swift
```diff
@@ -0,0 +1,53 @@
...
+    public var name: String {
+        return String(self)
+    }
+
+}
+
+// MARK: - FileType
+
+public enum FileType: String, Equatable {
+    case h, json, m, markdown = "md", pbxproj, plist, storyboard, swift
```
SD10: What do you think about adding the extensions for `.xcworkspace`, `.xcscheme`, `.yml` and `.mm`?

SD10: Yep, I think that's a good idea 👍

orta:
`Error: Error: Results passed to Danger JS did not include fails` I feel is a weird thing we've been seeing in danger-js, I don't think it's your issue

orta:
Perfect 👍

orta:
Danger JS is less mature, that's all. That bug has been real hard to get in danger-js.

FWIW, I'll never build danger for Swift myself, too small of a community (and largely inactive in contributing back) for such a large amount of my time, this technique allows danger-swift to actually exist and to potentially allow it to run on Peril.

yhkaplan:
Fair enough. 👍

PR #84, opened 2018-10-07
Supports using the new url option for sub-processes

orta:
This should fix https://github.com/danger/danger-swift/issues/69 - but it requires shipping Danger JS 5.0 

For now this is both backwards and future compatible on both sides, as it does both the url and the json and Danger JS supports listening for both the url and the JSON (but prioritizes the URL)

Docs will be shipped on the danger systems page for `danger process`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ matrix:
       osx_image: xcode10
       install:
         - node -v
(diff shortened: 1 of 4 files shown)


orta:
Danger 5.0 is shipped

PR #90, opened 2018-10-25
Restored danger-swift edit functionality

f-meloni:


orta:
Hah, yep, OK, I'll ship a release right away

PR #91, opened 2018-10-25
Added Shellout files on the Makefile

f-meloni:
If you try to run `danger-swift pr $PR` and you are using the installed `swift-danger` (then you don't full `.build` folder) you get this error.

```sh
Danger: ⅹ Failing the build, there is 1 fail.

Dangerfile.swift:4:8: error: missing required module 'ShellOut'
import Danger
       ^
```

Not sure this is the best solution to the problem, but adding this and calling `make install` solved the problem in my case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ## Master
 
+* Added Shellout files on the Makefile [#91](https://github.com/danger/danger-swift/pull/91) by [f-meloni](https://github.com/f-meloni)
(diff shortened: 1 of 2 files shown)


f-meloni:
Looks is having the same problem of https://github.com/danger/danger-swift/pull/89

f-meloni:
Now the problem is different, travis is not working anymore due the repo name change :D

orta:
Interesting, I added some more debug logs to this repo and it's a tad worrying that it just doesn't echo anything :-/

Issue #49, opened 2017-12-20
Support importing another file


orta:
This one is interesting, in JS/Ruby it's trivial to import another file - how can this happen with Danger?

orta:
Could add a unique require syntax which is pulled out before being passed to marathon, then the script can reference the others? That could then work with `danger edit` too.

PR #93, opened 2018-11-03
Multiple files import

f-meloni:
Opening this even if i still have to think to how to apply the canges to `danger-swift edit` just to get some early feedbacks, because the taken approach is different to the one suggested on https://github.com/danger/swift/issues/49.

### DIFF
--- a/Dangerfile.swift
+++ b/Dangerfile.swift
@@ -5,22 +5,10 @@ import Danger
 
 let danger = Danger()
 
-let allSourceFiles = danger.git.modifiedFiles + danger.git.createdFiles
-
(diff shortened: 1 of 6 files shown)


orta:
Interesting! 

Yeah, this looks 👍 to me

orta:
OK, I need to try figure out what's going on with these 10m launches :-/

f-meloni:
Thanks :)
Will add the changelog in the meantime, to remove the warning

orta:
interesting, none of the logs got hit

orta:
I wonder if this is because it's using the homebrew version of danger-js

PR #94, opened 2018-11-05
Improve danger swift edit

f-meloni:
This might be a little bit controversial in some points, i will try to put as much as possible details, images and inline comments about what i'm trying to do

## PR Goal

screenshot 2018-11-05 at 19 43 40

If you run `danger-swift edit` you will see that, with no syntax highlighting, no auto competition.
The reason for that was that it was not correctly overriding the `OTHER_SWIFT_FLAGS`.
But once i did override them correctly (was enough remove the `,` `let before = "-DXcode\","` to fix that) the syntax highlighting was correct, but it was not compiling.
In order to make it compile i had to link the danger library, this is why i though that was a good idea to actually use the `--xcconfig-overrides` option on the `generate-xcodeproj` command

This is the result

screenshot 2018-11-05 at 19 46 59

### DIFF
--- a/Sources/Danger/Danger.swift
+++ b/Sources/Danger/Danger.swift
@@ -26,6 +26,12 @@ private final class DangerRunner {
         logger.logInfo("Ran with: \(CommandLine.arguments.joined(separator: " "))")
         
         let cliLength = CommandLine.arguments.count
+        
+        guard cliLength - 2 > 0 else {
+            logger.logError("To execute Danger run danger-swift ci , danger-swift pr or danger-swift local on your terminal")
+            exit(1)
+        }
+        
         let dslJSONArg: String? = CommandLine.arguments[cliLength - 2]
         let outputJSONPath = CommandLine.arguments[cliLength - 1]
 
--- a/Sources/Danger/Danger.swift
+++ b/Sources/Danger/Danger.swift
@@ -65,72 +65,142 @@ public func Danger() -> DangerDSL {
     return DangerRunner.shared.dsl
 }
 
+extension DangerDSL {
+    /// Fails on the Danger report
+    public var fails: [Violation] {
+        return DangerRunner.shared.results.fails
+    }
+    
+    /// Warnings on the Danger report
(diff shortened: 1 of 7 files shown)


Sources/Danger/Danger.swift
```diff
@@ -26,6 +26,12 @@ private final class DangerRunner {
         logger.logInfo("Ran with: \(CommandLine.arguments.joined(separator: " "))")
         
         let cliLength = CommandLine.arguments.count
+        
+        guard cliLength - 2 > 0 else {
```
f-meloni: Now is possible to run the code, but if you run it crashes

screenshot 2018-11-05 at 20 01 02

This is why i added a nice message that replaces the crash

Sources/Danger/Danger.swift
```diff
@@ -65,72 +71,142 @@ public func Danger() -> DangerDSL {
     return DangerRunner.shared.dsl
 }
 
+extension DangerDSL {
+    /// Fails on the Danger report
+    public var fails: [Violation] {
```
f-meloni: I added this to allow people that are using multiple files to use danger to post messages etc. with typing and auto complete, without have to `import Danger` in all the files (given you have a global instance of danger defined on the Dangerfile, and that the multiple files will be copied on the Dangerfile when danger is executed)
(It works also with the import anyway)

But in this way you get 
screenshot 2018-11-05 at 19 48 07
without import Danger

Sources/Runner/Script+Imports.swift
```diff
@@ -51,6 +59,24 @@ extension Script {
         let fileName = importPath.split(separator: "/").last
         return folder.path + "Sources/\(name)/\(fileName ?? "")"
     }
+    
+    private func generateXCodeProjWithConfig(configPath: String) throws {
```
f-meloni: Would be better if we could pass to Marathon the xcconfig for the xcodeproj in the arguments, maybe worth propose that on the Marathon repo

orta:
OK, yeah, I think this idea is 👍

Issue #17, opened 2017-09-02
Danger will eventually need a GitHub API client


orta:
Existing libs:

* https://github.com/onmyway133/Github.swift
* https://github.com/nerdishbynature/octokit.swift

Both, very reasonably don't support SwiftPM and both have dependencies that would need to be resolved too. 

If you ask me, this is something that can easily be code-gen'd - the JS version we use in Danger JS is basically using this routes.json to generate all the different versions: https://github.com/mikedeboer/node-github/blob/master/lib/routes.json

So it may be easier to build that as a new SwiftM package than to port another and all its deps.

kdawgwilk:
I actually have a working GitHub client somewhere on my machine because I came across both of these as well and needed it for a bot I was writing in Swift.  I will try and dig it up

orta:
In the next Danger JS alpha the DSL will include a bunch of GitHub related metadata: https://github.com/danger/danger-js/pull/395/files#diff-7c330aa12e1d8f926998264b38ca39feR23 ( see the `DangerDSLJSONType` in `source/dsl/DangerDSL.ts` if the URL 404s. This should be enough metadata to bootstrap any API client.

orta:
Also, @kdawgwilk has an SPM fork of octokit.swift - https://github.com/kdawgwilk/octokit.swift which could get turned into something for production

eneko:
I've been looking at Tentacle a little bit this last week. It is Swift 4 and SPM-ready. 

I like how easy it looks, though haven't used it. However, I don't like that it depends on ReactiveSwift. It feels like a really _big_ dependency to bring, and large dependencies are no bueno :(

orta:
Yeah, agreed, [I wouldn't accept](http://artsy.github.io/blog/2015/09/18/Cocoa-Architecture-Dependencies/#Community.Relationships) a lib with a transitive on Reactive Swift as a dep into Danger.

SD10:
@orta If you get a minute can you elaborate on the code generation? I'd like to look more into this topic.

> If you ask me, this is something that can easily be code-gen'd - the JS version we use in Danger JS is basically using this routes.json to generate all the different versions: https://github.com/mikedeboer/node-github/blob/master/lib/routes.json

I wouldn't mind writing us our own API client free of dependencies. It seems like there are a few libraries that attempt to do this and then die off. Maybe if I throw it in an organization with Aeryn the community will help maintain it 😅

eneko:
[...]

PR #95, opened 2018-11-06
Github API Client

f-meloni:
This should solve https://github.com/danger/swift/issues/17


Dangerfile.swift
```diff
@@ -15,3 +15,18 @@ if (danger.git.createdFiles.count + danger.git.modifiedFiles.count - danger.git.
...
+    case .success(let user):
+        message(user.name ?? "")
+    case .failure:
+        break
+    }
+    
+    group.leave()
+}
+
+group.wait()
```
orta: cool, that worked, do we need the dispatch group? It should wait because danger waits till the end of the process

Sources/Danger/DangerDSL.swift
```diff
@@ -15,4 +16,32 @@ public struct DangerDSL: Decodable {
...
+        let settings = try container.decode(Settings.self, forKey: .settings)
+        
+        if github != nil {
+            let config: TokenConfiguration
+            
+            if let baseURL = settings.github.baseURL {
+                config = TokenConfiguration(settings.github.accessToken, url: baseURL)
+            } else {
+                config = TokenConfiguration(settings.github.accessToken)
+            }
```
orta: Cool, yep, this is what I expected 👍

Sources/Danger/GitHubDSL.swift
```diff
@@ -28,7 +29,8 @@ public struct GitHub: Decodable, Equatable {
     public let reviews: [GitHubReview]
 
     public let requestedReviewers: GitHubRequestedReviewers
-
+    
+    internal(set) public var api: Octokit!
```
orta: internal? ( maybe I don't get how access rights work here)

Tests/DangerTests/DangerDSLResources/DangerDSLGitHub.swift
```diff
@@ -0,0 +1,2126 @@
...
+              "avatar_url": "https://avatars3.githubusercontent.com/u/498212?v=4",
+              "gravatar_id": "",
+              "url": "https://api.github.com/users/ashfurrow",
+              "html_url": "https://github.com/ashfurrow",
+              "followers_url": "https://api.github.com/users/ashfurrow/followers",
+              "following_url": "https://api.github.com/users/ashfurrow/following{/other_user}",
+              "gists_url": "https://api.github.com/users/ashfurrow/gists{/gist_id}",
+              "starred_url": "https://api.github.com/users/ashfurrow/starred{/owner}{/repo}",
+              "subscriptions_url": "https://api.github.com/users/ashfurrow/subscriptions",
+
```
orta: ^ - this might have got revoked (check your emails) - unless it is one of mine?

Sources/Danger/GitHubDSL.swift
KITSFrancoMeloni: Just the set is internal (the api is set from DangerDSL, then it can not be `private(set)`)

Tests/DangerTests/DangerDSLResources/DangerDSLGitHub.swift
f-meloni: Is the one on the fixtures, but anyway is just to test it parses it correctly, then shouldn't matter

orta:
Looks good - want to add a changelog entry then you can merge 👍

PR #98, opened 2018-11-12
Allow swift run

orta:
This PR adds adds some docs, ensure that danger-swift can run via `run swift` and hides a lot of the default logging behind the debug/verbose flags.

### DIFF
--- a/Sources/Runner/main.swift
+++ b/Sources/Runner/main.swift
@@ -8,13 +8,15 @@ do {
     let logger = Logger(isVerbose: isVerbose, isSilent: isSilent)
 
     if cliLength > 1 {
-        logger.logInfo("Launching Danger Swift \(CommandLine.arguments[1]) (v\(DangerVersion))")
+        logger.debug("Launching Danger Swift \(CommandLine.arguments[1]) (v\(DangerVersion))")
         switch(CommandLine.arguments[1]) {
         case "ci", "local", "pr":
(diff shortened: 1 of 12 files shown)


Sources/Runner/main.swift
```diff
@@ -8,13 +8,15 @@ do {
...
     if cliLength > 1 {
-        logger.logInfo("Launching Danger Swift \(CommandLine.arguments[1]) (v\(DangerVersion))")
+        logger.debug("Launching Danger Swift \(CommandLine.arguments[1]) (v\(DangerVersion))")
         switch(CommandLine.arguments[1]) {
         case "ci", "local", "pr":
             let exitCode = try runDangerJSCommandToRunDangerSwift(CommandLine.arguments[1], logger: logger)
             exit(exitCode)
         case "edit":
             try editDanger(logger: logger)
+        case "runner":
```
f-meloni: Should this be `run`?

orta: not in this case, runner makes it the same as danger-js - it represents executing the runtime/dangerfile which right now `danger-swift` represents (eventually I'll switch that off)

f-meloni: Yes i figured that out, just personally preferred `danger-swift run jsonIn jsonOut` would nicer than `danger-swift runner jsonIn jsonOut`.
In any case i think this is a great idea, that would allow us to test it easly :)

Issue #96, opened 2018-11-07
Support custom dangerfile


marcelofabri:
Using `--dangerfile` would be great to split multiple setup into dangerfiles (e.g. one for PR validation and one that needs build results)

This is supported on danger-js and danger-ruby already.

orta:
Makes sense, yeah

f-meloni:
@orta Is it possible to pass the Dangerfile's custom path, if any, on the danger DSL settings object? (Given the parameter `dangerfile` is passed to `danger-js` in any case when you call `danger-swift ci`, `local` or`pr`)
Given the current main implementation I would rather receive it on the JSON than as parameter 
Does it looks a good approach to you?

orta:
Yeah, I think that's a real good thing to do  - it might get a bit tricky with the commands like `danger-swift ci` which will need documenting as supporting them too, and right now danger-js will validate that the args passed are pre-defined (and will bail if it isn't supported, this can be turned off though) 

👍

orta:
Fixed by #100

PR #100, opened 2018-11-13
Dangerfile arg support

f-meloni:
This solves https://github.com/danger/swift/issues/96

One of the things i would to do (and will probably do soon), is create a different lib that is used by the Runner (something like `RunnerLib`), and put there all the classes that are now on Danger lib but are useful just for the Runner.

### DIFF
--- a/Sources/Danger/CliArgs.swift
+++ b/Sources/Danger/CliArgs.swift
@@ -1,0 +1,51 @@
+//
+//  CliArgs.swift
+//  Danger
+//
+//  Created by Franco Meloni on 12/11/2018.
+//
+
+// Describes the possible arguments that could be used when calling the CLI
+public struct CliArgs {
+    // So you can have many danger runs in one code review
+    public let id: String?
+    
+    // The base reference used by danger PR (e.g. not master)
+    public let base: String?
+    
+    // For debugging
+    public let verbose: String?
+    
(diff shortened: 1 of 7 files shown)


Sources/Danger/CliArgs.swift
```diff
@@ -0,0 +1,51 @@
...
+    fileprivate enum CodingKeys: String {
+        case id
+        case base
+        case verbose
+        case externalCiProvider
+        case textOnly
+        case dangerfile
+    }
+    
+    init(dictionary: [String:Any]) {
```
f-meloni: I didn't want to make this part of the parsed Danger struct - or of the Danger runner (because on the `Dangerfile` you probably don't need this, and in the runner you don't need all the other `Danger` related data).
Then didn't make a lot of sense to me to parse the entire DSL with `JSONEncoder`, i preferred to use this approach.
Also `JSONSerialization` is more performant than `Codable`, and given i will have to parse JSON two times (one on the Dangerfile and one on the Runner), i preferred to use `JSONSerialization` than `Codable` in this case

orta: This makes sense to me 👍

orta: and also protects against it changing and adding new attributes in the future

PR #103, opened 2018-11-14
Run all the tests on linux

f-meloni:
Given we are close to the v1.0 i think is time to do a little bit of code cleaning and improvements :)

### DIFF
--- a/Tests/DangerTests/FileTests.swift
+++ b/Tests/DangerTests/FileTests.swift
@@ -1,7 +1,7 @@
 import XCTest
 @testable import Danger
 
-class FileTests: XCTestCase {
+final class FileTests: XCTestCase {
     
     static var allTests = [
         ("test_fileType_forHFile", test_fileType_forHFile),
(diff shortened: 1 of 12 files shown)


Tests/DangerTests/FileTests.swift
```diff
@@ -137,4 +138,13 @@ class FileTests: XCTestCase {
...
     }
+    
+    func testLinuxTestSuiteIncludesAllTests() {
+        #if !os(Linux)
+        let thisClass = type(of: self)
+        let linuxCount = thisClass.allTests.count
+        let darwinCount = thisClass.defaultTestSuite.tests.count
+        XCTAssertEqual(linuxCount, darwinCount, "\(darwinCount - linuxCount) tests are missing from allTests")
+        #endif
+    }
```
orta: hah - this is great

Tests/DangerTests/GitTests.swift
```diff
@@ -1,14 +1,17 @@
 import XCTest
 import Danger
 
-class GitTests: XCTestCase {
+final class GitTests: XCTestCase {
```
orta: does this final really add any value?

f-meloni: Even if minimum it should improve the performances because the function call doesn't go through the virtual dispatch.

Issue #104, opened 2018-11-14
Call "danger-js" to trigger danger-js to do any work instead of "danger"


orta:
This was added in danger-js 6.1.3

PR #105, opened 2018-11-14
 Use danger-command for calls instead of danger command

f-meloni:
Fixes https://github.com/danger/swift/issues/104 by calling`danger-command`

e.g.
it calls 
`danger-ci --process danger-swift` when you call`danger-swift ci`


orta:
Ace 👍

PR #109, opened 2018-11-16
Separate the Danger library from the Runner

f-meloni:
Originally some stuff that was used only from the runner was on the Danger library.
I thought it was a good idea to actually make them independent, and also it was cleaner to have a specific library for the stuff we use on the Runner


orta:
Cool, seems fine with me 👍

PR #110, opened 2018-11-17
Suggestions

f-meloni:
This is just until it doesn't get implement on danger-js


Sources/Danger/Danger.swift
```diff
@@ -206,6 +218,11 @@ public func markdown(message: String, file: String, line: Int) {
     DangerRunner.shared.dsl.markdown(message: message, file: file, line: line)
 }
 
+/// Adds an inline suggestion to the Danger report
+public func suggestion(code: String, file: String, line: Int) {
```
DangerCI:
```suggestion
Test Suggestion
```

orta:
Cool, yeah, looks good 👍

PR #116, opened 2018-11-21
Attempt at grabbing the DSL async and with a timeout

orta:
This isn't working yet, but better to have WIP PR


orta:
sigh, didn't fix  the core issue, but did not that it got past grabbing stdin

orta:
It's possible that #128 fixed this, but I'm not sold.

PR #122, opened 2018-11-23
Help

f-meloni:
this is the help result from terminal.

screenshot 2018-11-23 at 10 53 40

Also hopefully it auto generates the LinuxMain now and we won't have to maintain it anymore :D


f-meloni:
`swift test --generate-linuxmain` can be executed just on osx because requires objc runtime, but is still better than maintain the list of tests manually.
I've added also another instance of travis that checks that the tests are always up to date in the files that are used by Linux

orta:
Nice - this should be `danger-swift --help` and not a separate command,  this allows all commands to have a help command - other than that - looks 👍

Sources/RunnerLib/HelpMessagePresenter.swift
```diff
@@ -0,0 +1,23 @@
...
+//  HelpMessagePresenter.swift
+//  RunnerLib
+//
+//  Created by Franco Meloni on 23/11/2018.
+//
+
+import Logger
+
+public final class HelpMessagePresenter {
+    public static func showHelpMessage(command: DangerCommand?, logger: Logger) {
```
f-meloni: I passed the command here then we will be able to show the command specific help if any

orta: 👍

f-meloni:
merge on green

PR #123, opened 2018-11-23
Adds a utils set of funds, starting with one for reading a text file from the disk

orta:
I spotted we have a few copyright headers and figured we should remove them for consistency, but better to make a rule to stop them upfront. In making the rule I added a new util function.


Sources/Danger/DangerUtils.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+
+/// Utility functions that make Dangerfiles easier to write
+
+struct DangerUtils {
```
f-meloni: This should be public

Sources/Danger/DangerDSL.swift
```diff
@@ -16,7 +16,9 @@ public struct DangerDSL: Decodable {
     public let github: GitHub!
 
     public let bitbucket_server: BitBucketServer!
-    
+
+    public let utils: DangerUtils!
```
f-meloni: i think the `!` is not needed

orta:
These would have been caught by Komondor!

PR #126, opened 2018-11-25
Adds Komondor, SwiftLint and SwiftFormat

orta:
This adds, and automates using SwiftLint and SwiftFormat via Komondor. 

So the diff is pretty big, I'll highlight the stuff I actually did.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,8 +28,10 @@ uninstall:
 deploy:
 	if [[ -z "$(NEW_VERSION)" ]]; then echo "Please add a value to the NEW_VERSION variable"; exit 1; fi
 	`Scripts/update_makefile.sh`
+	`Scripts/dev_deps_hide.rb`
 	`Scripts/update_danger_version.sh`
 	`Scripts/update_changelog.sh`
 	make docs
 	`Scripts/commit_new_version.sh`
 	`Scripts/create_homebrew_tap.sh`
+	`Scripts/dev_deps_unhide.rb`
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:4.0
+// swift-tools-version:4.2
 
 import PackageDescription
 
@@ -27,8 +27,7 @@ let package = Package(
         .target(name: "Runner", dependencies: ["RunnerLib", "MarathonCore", "Logger"]),
         .testTarget(name: "DangerTests", dependencies: ["Danger"]),
         .testTarget(name: "RunnerLibTests", dependencies: ["RunnerLib"]),
-    ],
-    swiftLanguageVersions: [4]
+    ]
 )
 
 #if canImport(PackageConfig)
--- a/Package.swift
+++ b/Package.swift
@@ -8,14 +8,17 @@ let package = Package(
     name: "danger-swift",
     products: [
         .library(name: "Danger", type: .dynamic, targets: ["Danger"]),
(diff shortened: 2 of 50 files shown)


Makefile
```diff
@@ -28,8 +28,10 @@ uninstall:
 deploy:
 	if [[ -z "$(NEW_VERSION)" ]]; then echo "Please add a value to the NEW_VERSION variable"; exit 1; fi
 	`Scripts/update_makefile.sh`
+	`Scripts/dev_deps_hide.rb`
 	`Scripts/update_danger_version.sh`
 	`Scripts/update_changelog.sh`
 	make docs
 	`Scripts/commit_new_version.sh`
 	`Scripts/create_homebrew_tap.sh`
+	`Scripts/dev_deps_unhide.rb`
```
orta: I added these two extra steps to the deploy, ruby because everyone has ruby installed by default

Package.swift
```diff
@@ -8,14 +8,17 @@ let package = Package(
...
     dependencies: [
         .package(url: "https://github.com/JohnSundell/Marathon.git", from: "3.1.0"),
         .package(url: "https://github.com/JohnSundell/ShellOut.git", from: "2.1.0"),
         .package(url: "https://github.com/nerdishbynature/octokit.swift", from: "0.9.0"),
         // Dev dependencies
-        .package(url: "https://github.com/eneko/SourceDocs.git", from: "0.5.1"),
+        .package(url: "https://github.com/eneko/SourceDocs.git", from: "0.5.1"), // dev
+        .package(url: "https://github.com/orta/Komondor.git", from: "1.0.0"), // dev
+        .package(url: "https://github.com/nicklockwood/SwiftFormat.git", from: "0.35.8"), // dev
+        .package(url: "https://github.com/Realm/SwiftLint.git", from: "0.28.1"), // dev
```
orta: If a dep has "// dev" after it, then it will be commented out

f-meloni: Cool, could it just comment everything that is after `// Dev dependencies`?

Package.swift
```diff
@@ -24,6 +27,21 @@ let package = Package(
...
+    ]
 )
+
+#if canImport(PackageConfig)
+    import PackageConfig
+
+    let config = PackageConfig([
+        "komondor": [
+            "pre-push": "swift test",
+            "pre-commit": [
```
f-meloni: Would be cool to add here
`swift test --generate-linuxmain`

Scripts/commit_new_version.sh
```diff
@@ -1,9 +1,6 @@
 #!/bin/bash
 
-git add Makefile
-git add CHANGELOG.md
-git add Sources/Runner/main.swift
-git add Documentation
+git add Makefile CHANGELOG.md Sources/Runner/main.swift Documentation Package.swift
```
f-meloni: 👍

Package.swift
f-meloni: something like: everything that starts with `.package(url:` after `// Dev dependencies`

Package.swift
f-meloni: Then we can remove the travis thing that checks if the checks are synced, or just put `swift test --generate-linuxmain` after the swiftlint call to avoid that it fails, and keep it there, because it could still fail if a person didn't call the `komondor init`

orta:
Merging from master helped. I'm gonna keep the `// dev` because it makes the script very simple, and easy to copy to other projects, I think it's a good idea though if it weren't a one-off

PR #127, opened 2018-11-25
Don't use readDataToEndOfFile to read the JSON to avoid it hangs there forever

f-meloni:


f-meloni:
Didn't work :( closing

PR #128, opened 2018-11-25
Use a url to pass the JSON DSL instead

orta:
looks like some other stuff got included, will deal with that later


orta:
Eh, let's give it a shot and see where it goes

PR #129, opened 2018-11-26
Fix Bitbucket Server parsing

f-meloni:
Given it never worked I did some breaking changes to the public interface to make it consistent with itself and more swifty


orta:
👍

f-meloni:
merge on green

PR #130, opened 2018-11-27
Use Logger as external dependency

f-meloni:


f-meloni:
merge on green

f-meloni:
merge on green

PR #132, opened 2018-11-30
Don't send the command again to danger-js

f-meloni:
Currently we are sending the same command two times to danger-js 

e.g.
```
danger-pr pr --process danger-swift PRLink
```


f-meloni:
merge on green

PR #134, opened 2018-12-03
Update swiftformat

f-meloni:


f-meloni:
merge on green

Package.swift
```diff
@@ -39,7 +39,7 @@ let package = Package(
             "pre-commit": [
                 "swift test",
                 "swift test --generate-linuxmain",
-                "swift run swiftFormat .",
+                "swift run swiftformat .",
```
orta: Awesome, this mean vscode support is in 👍

PR #136, opened 2018-12-03
Add Rocket to release danger-swift

f-meloni:
Introducing https://github.com/f-meloni/Rocket to release the app.
I had few tests before open the PR but i will test it after this is merged with a new 0.7.3 release


.rocket.yml
```diff
@@ -0,0 +1,31 @@
+---
```
orta: PackageConfig please! (sorry)

f-meloni: Np i thought was too complex for it :) will move it 👍

orta: Dog fooding mainly, also open to helping figure out how to reduce that complexity either in PackageConfig, or Rocket

orta: For example, auto running `scripts/deploy/before.sh`,  `scripts/deploy/after.sh` if it exists etc 

We could steal some ideas form https://www.npmjs.com/package/release-it#command-hooks and get it so Rocket deploys with no config, and maybe Danger needs some because there are more parts

Package.swift
```diff
@@ -44,5 +45,22 @@ let package = Package(
                 "git add .",
             ],
         ],
+        "rocket": [
+            "steps": [
```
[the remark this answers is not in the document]
orta: nah, this is a great start 👍

PR #138, opened 2018-12-04
Fix htmlUrl documentation string

f-meloni:
`SourceDocs` was not generating the correct documentation for `htmlUrl` because there was one missing `/`


f-meloni:
merge on green

Issue #108, opened 2018-11-15
Add a timeout for waiting to STDIN?


orta:
Still trying to figure out why builds like https://travis-ci.org/danger/swift/jobs/455599717 are failing, maybe it's because STDOUT isn't being read and we're blocking on waiting for that? Danger JS is definitely sending it.

orta:
danger-js should now be able to send the DSL down, will just need to add a timeout and ship it in danger/swift

orta:
This is https://github.com/danger/swift/issues/108

orta:
Actually 
https://github.com/danger/swift/issues/97

PR #140, opened 2018-12-16
Improve travis caching, and improve the logging from Marathon

orta:
This would have made us spot #108 much quicker


orta:
merge on green

PR #143, opened 2018-12-16
Explore auto-generating the Swift interface for Danger

orta:
Uses the techniques in https://github.com/jpsim/SourceKitten/issues/405 to generate the Swift module interface directly, this matches https://danger.systems/js/reference.html / https://danger.systems/reference.html better.


Scripts/data/request.template.yml
```diff
@@ -0,0 +1,77 @@
+# This file
```
f-meloni: This should be removed i think

orta: ah yeah, I was meant to finish that

f-meloni:
should this maybe be added to rocket?

orta:
Yeah, probably, I'm not sure of this is the end result or a partial one for now, so I'm happy to let it be a bit out of date right now as I figure out docs.

PR #144, opened 2018-12-16
Remove additional build on travis.yml

f-meloni:
During the CI danger swift is already builded for test by `swift test` and builded again (for release) from `make install`, then i think we can save some time by avoid building it again with swift build


orta:
Yep, merge on green

f-meloni:
Looks like is taking one min less (total, not per build) 👍

PR #146, opened 2018-12-17
Merges in Danger SwiftLint

orta:
Closes https://github.com/ashfurrow/danger-swiftlint/issues/17 and merges SwiftLint into Danger. This will need another docs refresh, but then I think this'll get a big mention on the index for the website. 

I mainly brought it over wholesale, and then edited the tests to re-use some of the [new testing infra](https://github.com/danger/swift/blob/master/Documentation/usage/extending_danger_two.html.md) 

It'll need to support running via SwiftPM so we can use it on Danger Swift itself, but getting there.

This is a big enough move that I'd like to get sign off from both @ashfurrow @sunshinejr because moving this in re-assigns the copyright to the [Danger Swift license](https://github.com/danger/swift/blob/master/LICENSE) instead of the one from danger swiftlint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,22 @@
 
 ## Master
 
+- Whoah! The SwiftLint plugin has been merged into Danger by [@orta][]
+
+  This is not my work, really, it's the work of the contributors to the Danger SwiftLint plugin (which, yes, I am a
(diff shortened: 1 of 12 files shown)


orta:
OK, cool, let's get this in.

PR #147, opened 2018-12-18
Use Snapshots for testing

orta:
This [lovely library](https://github.com/pointfreeco/swift-snapshot-testing) works similar to how I write tests in JS land, which is useful, because I write a lot of tests there :D


Package.swift
```diff
@@ -19,10 +19,10 @@ let package = Package(
         .package(url: "https://github.com/nerdishbynature/octokit.swift", from: "0.9.0"),
         // Danger Plugins
         // Dev dependencies
-        .package(url: "https://github.com/eneko/SourceDocs", from: "0.5.1"), // dev
```
f-meloni: I deleted it by mistake :(

Sources/RunnerLib/Runtime.swift
```diff
@@ -31,7 +31,7 @@ public struct Runtime {
 
         if marathonScripts != nil {
             // TODO: Support running from a fork?
```
DangerCI: - :warning: TODOs should be resolved (Support running from a fork?). (`todo`)

PR #152, opened 2018-12-20
Use a fake printer to avoid warnings to be printed on tests

f-meloni:
If you run the tests you will see 

```
WARNING: File not fount at GeneredTestFile3.swift
```

This is because we are using the real logger there


f-meloni:
merge on green

PR #153, opened 2018-12-21
Fixes for people using via homebrew 

orta:
only the danger command is available, and not danger-pr etc because it only ships one exec

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,9 @@
 cache:
   directories:
-  # General SwiftPM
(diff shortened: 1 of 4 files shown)


orta:
This also adds a travis run to specifically run Danger via home-brew now, so we'd know if it fails in the future.

orta:
OK, the non-homebrew one works but it needs a deploy of this PR to make that work otherwise, so merging and will deploy.

PR #156, opened 2018-12-21
Fix danger report system

f-meloni:
`DangerResult` is a struct, on the old implementation it was copied, and the changes were not propagated to the original result object


f-meloni:
Yep working, the failing one is the homebrew one.
Going to release straight away.

Issue #131, opened 2018-11-30
ERROR: Failed to parse JSON again


phil-alekhin:
Hi guys!

I've faced with a similar problem like this https://github.com/danger/swift/issues/68 but I suppose that the reason is different.

Here is my DangerFile.swift

```swift
// Dangerfile.swift

import Danger
import DangerSwiftLint // package: https://github.com/ashfurrow/danger-swiftlint.git

SwiftLint.lint(inline: true, directory: "Project", configFile: ".swiftlint.yml", lintAllFiles: true, swiftlintPath: "./Pods/SwiftLint/swiftlint")
```

When danger-swiftlint trying to update packages I get an error. Here is an output:

```sh
Updating packages...
$ cd "/Users/speek/workspace/Speke-feed-iOS_PR-466-CAGLGXZWVRDBRDV4IDTYJ6D6L6X6JIBE7O2S6QPVBTCZKYZ7WWBA/" && /usr/bin/env xcrun --sdk macosx swift package --version
Apple Swift Package Manager - Swift 4.2.0 (swiftpm-14248)
$ cd "/Users/speek/.danger-swift/Packages/Generated/" && /usr/bin/env xcrun --sdk macosx swift package update
Fetching https://github.com/Speekly/danger-swiftlint.git
Fetching https://github.com/danger/danger-swift.git
Fetching https://github.com/JohnSundell/Files.git
Fetching https://github.com/JohnSundell/Marathon.git
Fetching https://github.com/JohnSundell/ShellOut.git
Fetching https://github.com/nerdishbynature/octokit.swift
Fetching https://github.com/eneko/SourceDocs.git
Fetching https://github.com/JohnSundell/Unbox.git
Fetching https://github.com/JohnSundell/Wrap.git
Fetching https://github.com/JohnSundell/Require.git
Fetching https://github.com/JohnSundell/Releases.git
Fetching https://github.com/nerdishbynature/RequestKit.git
Fetching https://github.com/jpsim/SourceKitten.git
Fetching https://github.com/Carthage/Commandant.git
Fetching https://github.com/eneko/MarkdownGenerator.git
Fetching https://github.com/onevcat/Rainbow
Fetching https://github.com/thoughtbot/Curry.git
Fetching https://github.com/drmohundro/SWXMLHash.git
Fetching https://github.com/jpsim/Yams.git
Fetching https://github.com/Quick/Quick.git
Fetching https://github.com/antitypical/Result.git
[...]

PR #158, opened 2018-12-24
Fail if danger-js version is below the minimum supported version

f-meloni:
As discussed on https://github.com/danger/swift/issues/131 would be good if danger-swift would fail if the danger-js used version is not supported.
It would avoid problems like the one we had when we introduced `--passURLForDSL` argument, that was not supported before the 6.1.2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 ## Master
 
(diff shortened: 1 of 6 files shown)


f-meloni:
schermata 2018-12-24 alle 17 32 47

f-meloni:
merge on green

orta:
great call

PR #161, opened 2018-12-29
Adds the 1.0 changelog

orta:
Eh, I think it's time

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,25 @@
 
 ## Master
 
+## 1.0.0
+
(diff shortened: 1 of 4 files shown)


f-meloni:
Ok, let me update rocket :P then we can release 1.0.0 with the new version

PS. what about the GitHub client?

CHANGELOG.md
```diff
@@ -13,6 +13,25 @@
 
 ## Master
 
+## 1.0.0
```
f-meloni: One small thing, I think is better keep this on master for now, because rocket will put 1.0.0 here when you will release :)

f-meloni:
merge on green

PR #165, opened 2018-12-31
Use SPM swiftlint if available

f-meloni:
Given we have SPM as suggested install method for danger i thought would be cool using SPM for swiftlint as well.
Then `swift run swiftlint` is called by default if you call `Swiftlint.lint()` and you have `SwiftLint` dep on your `Pacakge.swift`


f-meloni:
well at least it works :D

f-meloni:
ATM Swiftlint is running only against Sources, i will have to figure out how to exclude stuff like `XCTestManifests.swift` from the lint, before make it run against all modified files

PR #166, opened 2019-01-02
Append rule id to the Swiftlint message

f-meloni:
This is really useful if you want to disable the rule.
Emulates the default that swiftlint text on xcode.
swiftlint

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 ## Master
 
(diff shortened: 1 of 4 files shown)


f-meloni:
merge on green

orta:
maybe make them code 

> This is too many chars (`line_length`)

f-meloni:
Yep i will open another PR with that change

PR #168, opened 2019-01-03
Add comment to explain what is swiftlintPath default value #trivial

f-meloni:
I wanted to explain to people what is going to be used if they don't specify the `swiftlintPath` on the swiftlint call.
This in my opinion is also a way to make people evaluate SPM as tool to import `Swiftlint` in the project if they were not thinking about it.


orta:
👍

f-meloni:
merge on green

PR #179, opened 2019-01-25
Fix failing Swiftlint path trimming

absolute-heike:
https://github.com/danger/swift/blob/master/Sources/Danger/Plugins/SwiftLint/SwiftLint.swift#L79

```swift
let currentPath = currentPathProvider.currentPath
violations = violations.map { violation in
    let updatedPath = violation.file.deletingPrefix(currentPath).deletingPrefix("/")
    var violation = violation
    violation.update(file: updatedPath)
    return violation
}
```

Deleting the current path would fail, because the current path would include a newline at the end of the path.

See issue: https://github.com/danger/swift/issues/176

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 ## Master
 
 - Replace codable where was not needed by [@f-meloni][] - [#177](https://github.com/danger/swift/pull/177)
+- Fix malformed Swiftlint inline paths by [@absolute-heike][] - [#176](https://github.com/danger/swift/pull/176)
 
 ## 1.1.0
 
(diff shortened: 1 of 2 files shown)


f-meloni:
Nice!
just thinking if is worth using https://developer.apple.com/documentation/foundation/filemanager/1410766-currentdirectorypath instead of call pwd

absolute-heike:
This sounds like a great idea! :)

f-meloni:
@absolute-heike great!
can you please update the CHANGELOG? :)

f-meloni:
merge on green

PR #180, opened 2019-01-30
Swiftlint should consider excluded paths in config file

absolute-heike:
Right now changes in e.g. Pods folder are also checked by Swiftlint, even though they are excluded in the local `swiftlint.yml` config file. This bug was fixed 8 days ago in https://github.com/realm/SwiftLint/pull/2574 (version 0.30.1).

This PR changes the handling to pass the filelist to `swiftlint` as an array and set `--force-exclude` so the config exludes are adhered to.

Calls to Swiftlint should look like this: `SCRIPT_INPUT_FILE_COUNT=2 SCRIPT_INPUT_FILE_0="Source/File1.swift" SCRIPT_INPUT_FILE_1="Source/File2.swift" swiftlint lint --quiet --use-script-input-files --force-exclude --reporter json`


Sources/Danger/Plugins/SwiftLint/SwiftLint.swift
```diff
@@ -59,19 +59,24 @@ extension SwiftLint {
...
             }
 
-            violations = files.filter { $0.hasSuffix(".swift") }.flatMap { file -> [SwiftLintViolation] in
-                var arguments = ["lint", "--quiet", "--path \"\(file)\"", "--reporter json"]
-                if let configFile = configFile {
-                    arguments.append("--config \"\(configFile)\"")
-                }
-                let outputJSON = shellExecutor.execute(swiftlintPath, arguments: arguments)
-                return makeViolations(from: outputJSON, failAction: failAction)
+            // swiftlint takes input files in the format of `SCRIPT_INPUT_FILE_COUNT=2 SCRIPT_INPUT_FILE_0="file1.swift" SCRIPT_INPUT_FILE_1="file2.swift" swiftlint lint`
```
DangerCI: - :warning: Line should be 120 characters or less: currently 170 characters (`line_length`)

Sources/Danger/Plugins/SwiftLint/SwiftLint.swift
```diff
@@ -59,19 +59,24 @@ extension SwiftLint {
...
+            for (index, file) in files.enumerated() {
+                inputFiles.append(" SCRIPT_INPUT_FILE_\(index)=\"\(file)\"")
             }
+
+            var arguments = ["lint", "--quiet", "--use-script-input-files", "--force-exclude", "--reporter json"]
+            if let configFile = configFile {
+                arguments.append("--config \"\(configFile)\"")
+            }
+
+            let outputJSON = shellExecutor.execute([inputFiles, swiftlintPath].joined(separator: " "), arguments: arguments)
```
DangerCI: - :warning: Line should be 120 characters or less: currently 124 characters (`line_length`)

Sources/Danger/Plugins/SwiftLint/SwiftLint.swift
```diff
@@ -59,19 +59,24 @@ extension SwiftLint {
             violations = makeViolations(from: outputJSON, failAction: failAction)
         } else {
             // Gathers modified+created files, invokes SwiftLint on each, and posts collected errors+warnings to Danger.
-            var files = danger.git.createdFiles + danger.git.modifiedFiles
+            var files = (danger.git.createdFiles + danger.git.modifiedFiles).filter { $0.hasSuffix(".swift") }
```
f-meloni: I think you can use $0.fileType, you can use https://github.com/artsy/eigen/blob/master/Dangerfile.swift#L16 as example

f-meloni:
There are some tests that needs to be updated and maybe created (I would like to keep the swiftlint code well tested), but in general looks good 👍 
This should also make it quite faster when you have Swiftlint on you Package.swift! 🎉

absolute-heike:
Yeah, I will fix the tests for Swiftlint 👍 

Gonna be interesting, because the tests heavily rely on the command just being swiftlint (or the swiftlintPath). But I suffixed the command with the Environment variables 😬

I'm thinking of extending the ShellExecutor protocol to support environment variables, as this would ease up a lot of the issues that we have with the tests 😉 

> This should also make it quite faster when you have Swiftlint on you Package.swift! 🎉
You mean, because only one instance of Swiftlint is started, instead of one per file? 😄

f-meloni:
> You mean, because only one instance of Swiftlint is started, instead of one per file? 😄

Yes :)

f-meloni:
@absolute-heike if i understood your problem correctly I suppose you can use contains instead of == to find the command and then check that the final string is prefixed with the correct variables, something like
```swift
command == "SCRIPT_INPUT_FILE_COUNT=2 SCRIPT_INPUT_FILE_0=\"expectedFile0\" SCRIPT_INPUT_FILE_0=\"expectedFile1\" swiftlint"
```

absolute-heike:
Yes I could do that. Do you prefer this approach?

The change to ShellExecutor would look kinda like this:
```
func execute(_ command: String, arguments: [String] = [], environmentVariables: [String] = []) -> String {
        let script = "\(environmentVariables.joined(separator: " ")) \(command) \(arguments.joined(separator: " "))"
...
```

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@
 
 ## Master
 
+- Fix excluded Swiftlint paths [@absolute-heike][] - [#180](https://github.com/danger/swift/pull/180)
```
HeEAaD: Note: When you add `[@absolute-heike] http://github.com/absolute-heike` at the end of the file then it will be linked/rendered properly.

Sources/Danger/Plugins/SwiftLint/SwiftLint.swift
```diff
@@ -59,19 +59,25 @@ extension SwiftLint {
...
             }
+
+            // swiftlint takes input files in the format:
+            // `SCRIPT_INPUT_FILE_COUNT=2 SCRIPT_INPUT_FILE_0="file1" SCRIPT_INPUT_FILE_1="file2" swiftlint lint`
+            var inputFiles = ["SCRIPT_INPUT_FILE_COUNT=\(files.count)"]
+            for (index, file) in files.enumerated() {
+                inputFiles.append("SCRIPT_INPUT_FILE_\(index)=\"\(file)\"")
+            }
+
+            let outputJSON = shellExecutor.execute(swiftlintPath, arguments: arguments, environmentVariables: inputFiles)
```
DangerCI: - :warning: Line should be 120 characters or less: currently 121 characters (`line_length`)

absolute-heike:
Found a bug. When there is no file to lint (e.g. you didn't change a swift file), Swiftlint is still executed and prints `No lintable files found at paths: ''`, hence danger-swift is failing cause it cannot parse this message as JSON.
- [ ] do not execute swiftlint, if no files should be linted
- [ ] test this behavior

f-meloni:
merge on green

PR #181, opened 2019-01-31
Use correct decoding classes for the bitbucket decoding

f-meloni:
I've noticed that the models we were using to decode the BitBucket Server JSON are wrong, then I updated the models and the fixtures


Sources/Danger/BitBucketServerDSL.swift
```diff
@@ -298,23 +286,42 @@ public struct BitBucketServerPR: Decodable, Equatable {
...
     /// People who have participated in the PR
-    public let participants: [BitBucketServerAuthor]
+    public let participants: [Participant]
 
     // MARK: - BitBucketServerAuthor
 
-    public struct BitBucketServerAuthor: Decodable, Equatable {
+    /// A user that is parecipating in the PR
+    public struct Participant: Decodable, Equatable {
+        enum CodingKeys: String, CodingKey {
```
DangerCI: - :warning: Types should be nested at most 1 level deep (`nesting`)

f-meloni:
Would be cool to have a bitbucket server could use on the CI to run danger-swift pr against 😄

Sources/Danger-Swift/Fake.swift
```diff
@@ -0,0 +1 @@
+
```

Sources/Danger/BitBucketServerDSL.swift
```diff
@@ -298,24 +290,43 @@ public struct BitBucketServerPR: Decodable, Equatable {
...
             case user
         }
 
         /// The BitBucket Server User
         public let user: BitBucketServerUser
     }
+
+    /// A user that reviewed the PR
+    public struct Reviewer: Decodable, Equatable {
+        enum CodingKeys: String, CodingKey {
```
sunshinejr: hmm, why did you add `CodingKeys` enum? isn't that one synthesized for you already? 🤔

f-meloni:
merge on green

PR #183, opened 2019-02-04
Leave SPM decide if build again or not the DangerDependencies

f-meloni:
When I wrote this code I added a check to avoid to re build the library if was already on `.build`
With that you can end up in some corner cases if you change something (you add a dependency for example), and you have the .build cached, because the library is already builded and it won't be builded again.
SPM should anyway avoid to rebuild it if is already correctly builded, then i will leave the SPM algorithm decide

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 ## Master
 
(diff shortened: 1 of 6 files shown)


f-meloni:
merge on green

PR #184, opened 2019-02-07
Add SwiftLint to Docker image

alvarhansen:
At the moment SwiftLint plugin assumes that swiftlint binary is already available through shell or it can be installed using SPM. But `danger/swift` releases don't include SwiftLint (it and other "dev" packages are commented out).


orta:
Agree 👍

PR #187, opened 2019-02-09
Updated README to include SPM to install plugins

f-meloni:


Sources/Danger-Swift/Fake.swift
```diff
@@ -0,0 +1 @@
+
```

f-meloni:
merge on green

PR #191, opened 2019-02-10
Set github as var to make danger compile with swift 5

f-meloni:
Given we set `github.api` here https://github.com/danger/swift/blob/master/Sources/Danger/DangerDSL.swift#L61 on Swift 5 is required for github to be a `var`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
 
 ## Master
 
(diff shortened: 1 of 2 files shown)


f-meloni:
merge on green

f-meloni:
@orta peril is still complaining :(

orta:
no worries, it'll probably happen a few times - the latest Peril runtime was a big change

f-meloni:
👍

PR #193, opened 2019-02-11
Support unregistered users on BitbucketServer

f-meloni:
Bitbucket doesn't send you some of the user details if the email that did the commit is not related to a registered user


f-meloni:
merge on green

417-72KI (approved):

PR #195, opened 2019-02-15
Add ability to run swiftlint in strict mode

FabioTacke:
Introduce a strict mode that treats warnings like errors in inline mode, i.e. fails the build when a warning occurrs.


Sources/Danger/Plugins/SwiftLint/SwiftLint.swift
```diff
@@ -91,7 +94,10 @@ extension SwiftLint {
         violations = violations.map { violation in
             let updatedPath = violation.file.deletingPrefix(currentPath).deletingPrefix("/")
             var violation = violation
-            violation.update(file: updatedPath)
+            violation.file = updatedPath
+            if strict {
+                violation.severity = .error
```
f-meloni: can you please add a specific test for this? This would (I think), also change how the violation gets displayed on the markdown table, in that case that should be tested as well 👍

PR #196, opened 2019-02-16
Refactor code to fix SwiftLint warnings

ChaosCoder:
This PR is a follow-up to #195. This makes the code more readable, fixes swiftlint warnings introduced by #195 and reverts some whitespace changes #195 introduced.

There should be no changes in behaviour.


Sources/Danger-Swift/Fake.swift
```diff
@@ -0,0 +1 @@
+
```

Sources/Danger/Plugins/SwiftLint/SwiftLint.swift
```diff
@@ -46,86 +46,115 @@ extension SwiftLint {
...
+                                failAction: (String) -> Void) -> [SwiftLintViolation] {
+        if let directory = directory {
+            arguments.append("--path \"\(directory)\"")
+        }
+
+        let outputJSON = shellExecutor.execute(swiftlintPath, arguments: arguments)
+        return makeViolations(from: outputJSON, failAction: failAction)
+    }
+
+    private static func lintModifiedAndCreated(danger: DangerDSL,
```
DangerCI: - :warning: Function should have 5 parameters or less: it currently has 6 (`function_parameter_count`)

Sources/Danger/Plugins/SwiftLint/SwiftLint.swift
```diff
@@ -46,86 +46,115 @@ extension SwiftLint {
...
+            """
+            markdownMessage += violations.map { $0.toMarkdown() }.joined(separator: "\n")
+            markdownAction(markdownMessage)
         }
 
         return violations
     }
 
+    private static func lintAll(directory: String?,
+                                arguments: inout [String],
```
f-meloni: Why do you need arguments to be `inout`?
You can make `arguments` a var inside the function, given you don't need the updated value back

f-meloni:
can you please fix or add a swiftlint:disable to `Sources/Danger/Plugins/SwiftLint/SwiftLint.swift#L125`? Thank you :)

Sources/Danger/Plugins/SwiftLint/SwiftLint.swift
ChaosCoder: You are absolutely right. Thank you for spotting this Xcode "extract method" relic. :)

PR #198, opened 2019-02-19
Import correctly the files when one of the files contains an emoji

f-meloni:
Swift tries to simplify the string.count by returning 1 for an emoji even if that is actually created with more than one `char`.
That means that we can not use count to adjust the range if there are emoticons in the string


Sources/RunnerLib/Files Import/DangerFileGenerator.swift
```diff
@@ -10,7 +10,7 @@ public final class DangerFileGenerator {
 
         importsRegex.enumerateMatches(in: content, options: [], range: NSRange(location: 0, length: content.count), using: { result, _, _ in
             // Adjust the result to have the correct range also after dangerContent is modified
-            guard let result = result?.adjustingRanges(offset: dangerContent.count - content.count) else { return }
+            guard let result = result?.adjustingRanges(offset: dangerContent.endIndex.encodedOffset - content.endIndex.encodedOffset) else { return }
```
DangerCI: - :warning: Line should be 120 characters or less: currently 149 characters (`line_length`)

Sources/RunnerLib/Files Import/DangerFileGenerator.swift
```diff
@@ -10,7 +10,7 @@ public final class DangerFileGenerator {
 
         importsRegex.enumerateMatches(in: content, options: [], range: NSRange(location: 0, length: content.count), using: { result, _, _ in
```
DangerCI: - :warning: Line should be 120 characters or less: currently 140 characters (`line_length`)

PR #201, opened 2019-02-22
Update copyright notice to 2019

Atul9:


f-meloni:
thank you :)

f-meloni:
Looking at https://github.com/danger/danger/blob/master/LICENSE i think it should be just 2019 :)

Atul9:
I think for https://github.com/danger/danger/blob/master/LICENSE it should be updated to 2015-2019, it states from when to which year the project is copyrighted. 
Like this one - https://github.com/danger/danger-js/blob/master/LICENSE

f-meloni:
Ok :) I will tag @orta, just to double check he is happy with the change, otherwise we can revert it 👍

orta:
Yeah, I don't really mind - strictly speaking licenses only need their initial year of creation but I'm happy for other folks to keep the numbers up to date

PR #202, opened 2019-02-23
Create imported files if not present on edit command

f-meloni:


f-meloni:
merge on green

f-meloni:
Peril? :(

Sources/Danger-Swift/Fake.swift
```diff
@@ -0,0 +1 @@
+
```
DangerCI: - :warning: Files should have a single trailing newline. (`trailing_newline`)

PR #204, opened 2019-02-24
Fix danger edit for SPM

f-meloni:
`swift run danger-swift edit` doesn't compile if SPM generates the `libDanger.dylib` without the `Danger.swiftmodule` instead of generate only the `libDangerDeps.dylib` with the `Danger.swiftmodule` as it does on this repo.
e.g.
schermata 2019-02-24 alle 19 21 08

This PR fixes that


f-meloni:
Peril's danger, don't be mad at me :(

Sources/RunnerLib/SPMDanger.swift
```diff
@@ -28,7 +30,15 @@ public struct SPMDanger {
         _ = try? executor.shellOut(command: "swift build --product \(depsLibName)")
     }
 
-    public var libImport: String {
-        return "-l\(depsLibName)"
+    public var libsImports: [String] {
+        let libsImport = ["-l\(depsLibName)"]
+
+        // The danger lib is not always generated, this mainly happens on the danger repo, where the DangerDeps library and Danger.swiftmodule are enough
```
DangerCI: - :warning: Line should be 120 characters or less: currently 153 characters (`line_length`)

PR #216, opened 2019-03-26
Fix danger-swift parameters filter

f-meloni:
The filter was removing the parameter before the option instead of the one after


f-meloni:
merge on green

Sources/RunnerLib/GetDangerJSPath.swift
```diff
@@ -3,7 +3,7 @@ import Logger
 import ShellOut
 
 public func getDangerCommandPath(logger: Logger, args: [String] = CommandLine.arguments, shellOutExecutor: ShellOutExecuting = ShellOutExecutor()) throws -> String {
```
DangerCI: - :warning: Line should be 120 characters or less: currently 165 characters (`line_length`)

PR #224, opened 2019-04-13
Danger executor

f-meloni:
We use `ShellOut` in some case and our executor in some others.
Given we have also some small performances problems, I thought would be a good idea to refactor the code to use just our internal shell executor


Sources/DangerShellExecutor/ShellExecutor.swift
```diff
@@ -19,14 +47,14 @@ internal class ShellExecutor {
         task.waitUntilExit()
 
         let data = pipe.fileHandleForReading.readDataToEndOfFile()
-        return String(data: data, encoding: String.Encoding.utf8)!
+        return String(data: data, encoding: String.Encoding.utf8)!.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines)
```
DangerCI: - :warning: Line should be 120 characters or less: currently 126 characters (`line_length`)

Sources/RunnerLib/GetDangerJSPath.swift
```diff
@@ -1,13 +1,13 @@
+import DangerShellExecutor
 import Foundation
 import Logger
-import ShellOut
 
-public func getDangerCommandPath(logger: Logger, args: [String] = CommandLine.arguments, shellOutExecutor: ShellOutExecuting = ShellOutExecutor()) throws -> String {
+public func getDangerCommandPath(logger: Logger, args: [String] = CommandLine.arguments, shellOutExecutor: ShellExecuting = ShellExecutor()) throws -> String {
```
DangerCI: - :warning: Line should be 120 characters or less: currently 159 characters (`line_length`)

orta:
👌  less is less

👍

f-meloni:
merge on green

Issue #221, opened 2019-04-10
`swift build` fails compiling DangerDSL.swift


dgcoffman:
I'm attempting to following [the getting started guide](https://danger.systems/swift/guides/getting_started.html#swift-pm) by running `swift build` and it fails with following output:
```
/.build/checkouts/swift/Sources/Danger/DangerDSL.swift:61:24: 
error: immutable value 'self.github' may only be initialized once
            github.api = Octokit(config)
                       ^
/.build/checkouts/swift/Sources/Danger/DangerDSL.swift:16:12: 
note: change 'let' to 'var' to make it mutable
    public let github: GitHub!
           ^~~
           var
```

About me
```
swift -version
Apple Swift version 5.0 (swiftlang-1001.0.69.5 clang-1001.0.46.3)

swift package tools-version
5.0.0

System Version: macOS 10.14.4
```

Package.swift
``` swift
// swift-tools-version:5.0

import PackageDescription

let package = Package(
    name: "Eigen",
    dependencies: [
        .package(url: "https://github.com/danger/swift.git", from: "0.8.0")
    ],
    targets: [
        // This is just an arbitrary Swift file in our app, that has
        // no dependencies outside of Foundation, the dependencies section
        // ensures that the library for Danger gets build also.
        .target(name: "eigen", dependencies: ["Danger"], path: "", sources: ["DangerRequiresThis.swift"]),
    ]
)
``` 
(this example Package.swift was introduced recently in [this PR](https://gitlab.com/danger-systems/danger.systems/commit/d628fef8ca0f3e72aa36e6fd511563f2844f8d89#6b0f4e17f37f5ea51f169e6e1eaa04a0151b17d1_0_38) by @orta)

DangerRequiresThis.swift
```
import Foundation
```
___

I hand-edited DangerDSL.swift to make `github` a `var`, which let `swift build` complete with output 
```
[3/3] Compiling Swift Module 'eigen' (1 sources)
```

but `swift run danger edit` now yields `error: no executable product named 'danger'`

Any ideas?

dgcoffman:
I see there are apparently newer instructions in the README https://github.com/danger/swift/blob/master/README.md#swift-package-manager-more-performant

I then tried Package.swift
``` swift
// swift-tools-version:5.0

import PackageDescription

let package = Package(
    name: "Stopwatch",
    products: [
        .library(name: "DangerDeps[Product name (optional)]", type: .dynamic, targets: ["DangerDependencies"]),
    ],
[...]

PR #226, opened 2019-04-17
Bumped Marathon min version

f-meloni:
As per https://github.com/danger/swift/issues/221, use the new Marathon version that makes `danger-swift edit` work


f-meloni:
merge on green

PR #233, opened 2019-05-12
Remove not needed coding keys

f-meloni:


f-meloni:
merge on green

Sources/Danger/GitHubDSL.swift
```diff
@@ -179,15 +179,6 @@ public struct GitHubUser: Decodable, Equatable {
...
-
-    enum CodingKeys: String, CodingKey {
-        case id
-        case name
-    }
-
-    // MARK: - Properties
-
     /// The UUID for the team.
     public let id: Int
```
DangerCI: - :warning: Variable name should be between 3 and 40 characters long: 'id' (`identifier_name`)

PR #237, opened 2019-05-20
Fix Danger crashing when PR has no description

### LINKED ISSUE #236: Danger fails on PRs with no body
```
ERROR: Failed to parse JSON: valueNotFound(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "danger", intValue: nil), CodingKeys(stringValue: "github", intValue: nil), CodingKeys(stringValue: "issue", intValue: nil), CodingKeys(stringValue: "body", intValue: nil)], debugDescription: "Expected String value but found null instead.", underlyingError: nil))
ERROR: Dangerfile eval failed at Dangerfile.swift
```

Explanation: on PRs with no body set, GitHub most likely returns `null` for the issue's body instead of an empty string, which Danger cannot parse through `Codable` probably because the `body` property is non-optional.

Whilst PRs with no description are not very useful, this tool should be resilient enough to handle such a case.

I'll try and come up with a fix for this and make a PR for it.

davdroman:
Fixes #236.


f-meloni:
Thank you for the PR :)

I had a quick look and this should be not nullable.
This is the DSL declaration
https://github.com/danger/danger-js/blob/9a03a1103b63ac4f26e48fa1e94a46940c6eb146/source/dsl/GitHubDSL.ts#L188

And we have many PRs with no body here and they all passed without problems, last one is https://github.com/danger/swift/pull/235.
Will have a look at this a little bit deeper today.

f-meloni:
The only strange thing looks the `Issue`, that is actually what is making danger crash in your log.
@orta do you have any advice on this? 
On the DSL the GitHubIssue looks having just labels https://github.com/danger/danger-js/blob/9a03a1103b63ac4f26e48fa1e94a46940c6eb146/source/dsl/GitHubDSL.ts#L120
But on the DSL looks having a lot more stuff https://github.com/danger/swift/blob/master/fixtures/eidolon_609.json#L367

orta:
The issue is a bigggg object, I just didn't bother typing it  but almost all uses of it are just to grab the label so I didn't type the rest

f-meloni:
Ok, my suggestion then would be:
- Let's make the body optional just on the issue, because apparently is there.
- Let's make add a specific test on the decoding to be sure we don't have a regression in the future

davdroman:
I’ll take care of this.

davdroman:
Just for ref, here's an example of an issue with a `null` body https://api.github.com/repos/davdroman/badonde/issues/96. I'll be testing changes on my local branch against this PR with `danger-swift pr https://github.com/davdroman/Badonde/pull/96`.

davdroman:
@f-meloni sorted. I didn't add specific coverage for GitHubPR decoding even though I also made its body nullable as it is [required](https://api.github.com/repos/davdroman/badonde/pulls/96). Let me know what you think.

davdroman:
@f-meloni ready!

f-meloni:
merge on green

f-meloni:
I think I did it too late :D

PR #238, opened 2019-05-21
Remove swift 4.0 CaseIterable implementation

f-meloni:
Since we support only from swift `4.2`, the ifdef was not needed anymore

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 ## Master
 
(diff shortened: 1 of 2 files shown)


f-meloni:
Thank you Łukasz :)
My next goal is also try Danger Swift on Moya.
I just need to finish DangerSwiftProse before that, but I'm at a good point :P

f-meloni:
merge on green

PR #239, opened 2019-05-21
Move danger to swift 5.0

f-meloni:


f-meloni:
making some tries to see if the bug on linux is now fixed

PR #252, opened 2019-06-27
Handle not existing user

f-meloni:
If the user is not registered on GitHub the JSON we receive has `{}` in the author field for the commit.
This maybe can be fixed at danger-js level to just return nil in that case maybe, but I didn't see any obvious way? (any idea @orta?)
For the moment I added a patch to the problem


f-meloni:
Failing because of https://github.com/danger/danger-js/issues/884

PR #253, opened 2019-06-28
Allow nil GitHubMilestone descriptions

smalljd:
### The Problem

When `danger-swift` is executed on a PR that has a milestone associated with it, the JSON parsing will fail if the milestone's `description` field is `nil`. JSON parsing fails with the following message:

```sh
ERROR: Failed to parse JSON: valueNotFound(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "danger", intValue: nil), CodingKeys(stringValue: "github", intValue: nil), CodingKeys(stringValue: "issue", intValue: nil), CodingKeys(stringValue: "milestone", intValue: nil), CodingKeys(stringValue: "description", intValue: nil)], debugDescription: "Expected String value but found null instead.", underlyingError: nil))
ERROR: Dangerfile eval failed at Dangerfile.swift
ERROR: Could not get the results JSON file at /var/folders/md/w98btl4n7qbg0xy73lwczl38jqj51g/T/danger-response.json

```

### Proposed Solution

Since a `description` is an optional field for milestones, so too should it be optional in the `GitHubMilestone` struct.


smalljd:
Sounds good! I initially had it as a separate test, will change it back and update the changelog. Thanks for taking a look 😁

f-meloni:
Looks like `danger-js` is broken :(
cc. @orta

f-meloni:
If you merge master back should pass the CI now 👍

f-meloni:
merge on green

smalljd:
Sorry for being slow on the master-merge, thanks for the help @f-meloni!

f-meloni:
No problem :)
Just wanted to quickly release a version, in case someone had the same problem you had :)

PR #254, opened 2019-07-01
Fix link to eigen Dangerfile

dskuza:
This pull request fixes an issue in the iOS app tutorial, where the link to the `eigen` project's `Dangerfile` was broken, as`eigen` was mis-spelled as `migen` :)


f-meloni:
Thank you :) 
merge on green

PR #257, opened 2019-07-11
Make BitBucketServerUser emailAddress optional

f-meloni:
Users of type "SERVICE" don't have an emailAddress


f-meloni:
merge on green

PR #264, opened 2019-08-23
Fix not to log in to shell twice

manicmaniac:
Currently `ShellExecutor` always invokes `$SHELL` with `-l` option but this causes unexpected reloading of .bash_profile, .zprofile, .bashrc, .zshrc and so on.

Basically `danger-swift` is designed as a command line tool and expected to run on login shell so `-l` option is unnecessary.

Especially, since .bash_profile or .zprofile is only loaded at login time, some people expects it to be loaded at only once in a login session.

Thus I removed `-l` option from `ShellExecutor`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@
 
 ## Master
 
+## 2.0.2
+
+- Fix not to log in to shell twice [@manicmaniac][] - [#264](https://github.com/danger/swift/pull/264)
+
 ## 2.0.0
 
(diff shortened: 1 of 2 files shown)


manicmaniac:
@f-meloni I've just added new change log entry for 2.0.2.

BTW it looks like 2.0.1 doesn't have its change log entry. Is it okay to skip the entry for 2.0.1?

f-meloni:
Thank you!
2.0.1 was released just to make Danger-Swift work with GitHub Actions, will add a changelog entry later :)

f-meloni:
Merge on green

PR #268, opened 2019-09-06
Remove Swiftlint from Dockerfile

f-meloni:
Danger-Swift takes quite a lot when runs on GitHub Actions
Schermata 2019-09-06 alle 22 48 12
Approximately more than half of that time it to clone swiftlint, that not everyone uses, and someone uses from SPM.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,15 +7,9 @@ LABEL "com.github.actions.description"="Runs Swift Dangerfiles"
 LABEL "com.github.actions.icon"="zap"
 LABEL "com.github.actions.color"="blue"
 
-ARG SWIFT_LINT_VER=0.30.1
-
 # Install nodejs
 RUN curl -sL https://deb.nodesource.com/setup_10.x |  bash -
 RUN apt-get install -y nodejs
-
-# install SwiftLint
-RUN git clone -b $SWIFT_LINT_VER --single-branch --depth 1 https://github.com/realm/SwiftLint.git _SwiftLint
-RUN cd _SwiftLint && git submodule update --init --recursive; make install
 
 # Install danger-swift globally
 RUN git clone https://github.com/danger/danger-swift.git _danger-swift


f-meloni:
Schermata 2019-09-06 alle 23 12 17

This is this branch, and with all the dev dependencies, then it could be quite faster

f-meloni:
@orta do you have an opinion about this?

f-meloni:
I saw that there is an option to have inputs on Actions, we could explore this later to re add in an optional way

orta:
Yeah, that's what I was going to recommend (using the input to determine it) 

I feel like _lots_ of people would use it with SwiftLint - maybe we could have two dockerfiles?

f-meloni:
Ok, I've done few tests, and looks like the env variables are set after the build, then it can not work unless I put it on the `ENTRYPOINT`, that is not a good idea.
@orta can you give me more details about what you were thinking with the two `Dockerfile`s?
Because I think we can have only one, but we can have another repo with a `Dockerfile` that contains swiftlint maybe?

orta:
It'd need to be a [special branch or tag](https://help.github.com/en/articles/about-actions#versioning-your-action) per release with a different dockerfile. Totally scriptable on a release.

f-meloni:
Amazing idea! Let's merge this then I will create something with Rocket.
We could keep two Dockerfile and make two different tags in two different comments and then restore everything before we restore the dependencies

litso:
@f-meloni Do you have any tips on how to install swiftlint on the docker image now that it is removed?

f-meloni:
@litso there are few ways of doing it, the main ones I can think:
- Change the Rocket configuration of this repo to replace the Dockerfile with another Dockerfile that has swiftlint, create a `$version-swiflint` tag and then release restore the Dockerfile.
Given we use Rocket to release that will make a release with a Dockerfile that contains swiftlint for each release we make
- You can create your own Dockerfile like https://github.com/f-meloni/danger-swift-swiftlint-action

PR #270, opened 2019-09-14
Optimise danger actions clone

f-meloni:
This PR reduces the Danger-swift build time of one additional minute.

Before:
Schermata 2019-09-14 alle 16 25 43

After:
Schermata 2019-09-14 alle 16 36 49


orta:
good move!

PR #271, opened 2019-09-15
Update Marathon to a revision that supports swift 5.1

f-meloni:
Marathon is deprecated, and there is no release for https://github.com/JohnSundell/Marathon/pull/209, then I'm ok for us pointing the commit with the fix until Marathon is not replaced


f-meloni:
merge on green

PR #280, opened 2019-09-30
Fix for inline empty Dictionary: Bitbucket Cloud

khoogheem:
This fixes the issue where the json payload for the Comment can have inline as an empty dictionary:

```
            "inline" : {

            },
```

vs

```
            "inline" : {
            "to" : null,
            "path" : "client\/XXX\/Info.plist",
            "from" : 132
            },
```

### DIFF
--- a/Sources/Danger/BitBucketCloud.swift
+++ b/Sources/Danger/BitBucketCloud.swift
@@ -193,7 +193,7 @@ extension BitBucketCloud {
     public struct Comment: Decodable, Equatable {
         public struct Inline: Decodable, Equatable {
             public let from: Int?
-            public let to: Int // swiftlint:disable:this identifier_name
+            public let to: Int? // swiftlint:disable:this identifier_name
             public let path: String
         }
 
--- a/Sources/Danger/BitBucketCloud.swift
+++ b/Sources/Danger/BitBucketCloud.swift
@@ -228,6 +228,22 @@ extension BitBucketCloud {
 
         /// The user that created the comment
         public let user: User
+        
+        public init(from decoder: Decoder) throws {
+            let container = try decoder.container(keyedBy: CodingKeys.self)
+            
+            self.content = try container.decode(Content.self, forKey: .content)
+            self.createdOn = try container.decode(Date.self, forKey: .createdOn)
+            self.deleted = try container.decode(Bool.self, forKey: .deleted)
+            self.id = try container.decode(Int.self, forKey: .id)
+            // protect from "inline": {}
+            self.inline = (try? container.decodeIfPresent(Inline.self, forKey: .inline)) ?? nil
+            self.type = try container.decode(String.self, forKey: .type)
+            self.updatedOn = try container.decode(Date.self, forKey: .updatedOn)
+            self.user = try container.decode(User.self, forKey: .user)
+            
+        }
+
     }
 }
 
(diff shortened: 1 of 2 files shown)


Sources/Danger/BitBucketCloud.swift
```diff
@@ -228,6 +228,22 @@ extension BitBucketCloud {
 
         /// The user that created the comment
         public let user: User
+
```
DangerCI: - :warning: Lines should not have trailing whitespace. (`trailing_whitespace`)

Sources/Danger/BitBucketCloud.swift
```diff
@@ -228,6 +228,22 @@ extension BitBucketCloud {
...
+            self.content = try container.decode(Content.self, forKey: .content)
+            self.createdOn = try container.decode(Date.self, forKey: .createdOn)
+            self.deleted = try container.decode(Bool.self, forKey: .deleted)
+            self.id = try container.decode(Int.self, forKey: .id)
+            // protect from "inline": {}
+            self.inline = (try? container.decodeIfPresent(Inline.self, forKey: .inline)) ?? nil
+            self.type = try container.decode(String.self, forKey: .type)
+            self.updatedOn = try container.decode(Date.self, forKey: .updatedOn)
+            self.user = try container.decode(User.self, forKey: .user)
+
```
DangerCI: - :warning: Lines should not have trailing whitespace. (`trailing_whitespace`)

Sources/Danger/BitBucketCloud.swift
```diff
@@ -228,6 +228,22 @@ extension BitBucketCloud {
 
         /// The user that created the comment
         public let user: User
+        
+        public init(from decoder: Decoder) throws {
+            let container = try decoder.container(keyedBy: CodingKeys.self)
+
```
DangerCI: - :warning: Lines should not have trailing whitespace. (`trailing_whitespace`)

Sources/Danger/BitBucketCloud.swift
```diff
@@ -228,6 +228,22 @@ extension BitBucketCloud {
...
+        
+        public init(from decoder: Decoder) throws {
+            let container = try decoder.container(keyedBy: CodingKeys.self)
+            
+            self.content = try container.decode(Content.self, forKey: .content)
+            self.createdOn = try container.decode(Date.self, forKey: .createdOn)
+            self.deleted = try container.decode(Bool.self, forKey: .deleted)
+            self.id = try container.decode(Int.self, forKey: .id)
+            // protect from "inline": {}
+            self.inline = (try? container.decodeIfPresent(Inline.self, forKey: .inline)) ?? nil
```
f-meloni: You can just make `path` optional too, and that prevents from `"inline": {}` as well, but also from all the combinations that you can have (only path, only from etc).
If, on the other hand, it either have only zero or all the variables, I would then make them all required, and keep this line, what are your thoughts about this?

khoogheem: Clearly from the json I printed out.. To was set to null.  so I don't think you can make them all required.

Decoding JSON is like a losing battle if you don't control it ;)

khoogheem: I would be ok I guess with having everything in `inline` be optional

f-meloni: Yeah looks a good option :)

f-meloni:
Remember also to update the changelog with the change, the link to this PR and to your account :)

Sources/Danger/BitBucketCloud.swift
khoogheem: ok.. so we rather just make all of the fields in inline optional vs the init right?

f-meloni: 👍🏻

PR #311, opened 2020-01-12
Remove update dockerfile script

f-meloni:
Now that there is no version anymore on the Dockerfile there is no need to update it


f-meloni:
merge on green

PR #314, opened 2020-01-23
Add top-level properties for child DSLs

417-72KI:
There are top-level functions for reporting(e.g. `warn`, `fail`).
I think it would be nicer that there will be also properties for DSLs(e.g. `git`, `github`)

## Example 
- before
```swift
let danger = Danger()
if danger.github.pullRequest.title.lowercased().contains("[wip]") { // ← Redundant `danger` calling.
    warn("PR is classed as Work in Progress") // ← Already in top-level
}
```

- after
```swift
if github.pullRequest.title.lowercased().contains("[wip]") { // ← No more `danger` variable!
    warn("PR is classed as Work in Progress")
}
```

### DIFF
--- a/Sources/Danger/BitBucketServerDSL.swift
+++ b/Sources/Danger/BitBucketServerDSL.swift
@@ -395,3 +395,7 @@ extension BitBucketServer {
         public let type: String?
     }
 }
+
+public var bitbucketServer: BitBucketServer! {
+    return DangerRunner.shared.dsl.bitbucketServer
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
 
 ## Master
 
+- Add top-level properties for child DSLs [@417-72KI][] - [#314](https://github.com/danger/swift/pull/314)
+
 ## 3.0.0
 
 - Danger dependencies manager [@f-meloni][] - [#303](https://github.com/danger/swift/pull/303)
--- a/Sources/Danger/BitBucketCloud.swift
+++ b/Sources/Danger/BitBucketCloud.swift
@@ -244,3 +244,7 @@ extension BitBucketCloud {
         public let comment: Comment?
     }
 }
+
+public var bitbucketCloud: BitBucketCloud! {
+    return DangerRunner.shared.dsl.bitbucketCloud
+}
--- a/Sources/Danger/DangerUtils.swift
+++ b/Sources/Danger/DangerUtils.swift
@@ -82,3 +82,7 @@ public struct DangerUtils {
         return try shellExecutor.spawn(command, arguments: arguments)
     }
 }
+
+public var utils: DangerUtils {
+    return DangerRunner.shared.dsl.utils
+}
--- a/Sources/Danger/GitDSL.swift
+++ b/Sources/Danger/GitDSL.swift
@@ -54,3 +54,7 @@ extension Git {
         public let url: String
     }
 }
+
+public var git: Git {
+    return DangerRunner.shared.dsl.git
+}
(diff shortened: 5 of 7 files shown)


Sources/Danger/BitBucketServerDSL.swift
```diff
@@ -395,3 +395,7 @@ extension BitBucketServer {
         public let type: String?
     }
 }
+
+public var bitbucketServer: BitBucketServer! {
+    return DangerRunner.shared.dsl.bitbucketServer
+}
```
danger-public: - :warning: File should contain 400 lines or less: currently contains 401 (`file_length`)

danger-public:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against 4fd26d60babf229996b5f218f85c1adec9316257
</p>

f-meloni:
As you pointed out testing this would be kind of a problem, we ship `DangerFixtures` library that creates custom version of Danger with the data you need for your test, and could not be clear or easy understand which of the calls you can test and how IMO, if we start having a global function as well.

We took a similar path before, for being in line with other `Danger`s are doing, for the `warn`/`message`/`fail` functions, that decision brought us to have https://github.com/danger/swift/blob/bf35768e32cce87fdf5ad5cc0100dd26d5c09e5a/Sources/Danger/Report.swift#L16-L40, and probably makes sense, especially because a lot of people arrives from different Danger version (mainly ruby), and I think this makes the transition easier, given you could in a lot of cases kind of copy paste your old Dangerfile into the new one.

Then I would personally not suggest to to take that route given the small advantage it seems giving in this specific case, but I'm open to a discussion.

417-72KI:
@AvdLee @f-meloni 
Thanks for comments. I have some questions.
- Are you concerned that using top-level properties won't mock DSLs?
- Is `DangerFixtures` expected to be used by users?

I'm not clear for testing Danger, but IMHO `Danger Runner` itself should be mock in test.

orta:
My intention with this choice originally was so that it would be consistent with Danger JS (and rust, and Kotlin, and Python...) - not that it would be consistent with Danger Ruby. 

I'd rather it stays that way, personally.

417-72KI:
@orta Thanks, I understand and agree that.

In that case, my opinion is that to remove from top-level and always should be called via danger object.
I know it is redundant, but looks nature in Swift.

```swift
let danger = Danger()
if danger.github.pullRequest.title.lowercased().contains("[wip]") {
    danger.warn("PR is classed as Work in Progress")
}
```

orta:
[Danger JS looks a bit like this](https://www.typescriptlang.org/v2/en/play#code/JYWwDg9gTgLgBAbzgEwIYDsDmBTKAaOEbAZ2NRwJFSgGtkIB3dAh65uAM1WABs4BfTlAgg4AIjRZcYgLAAoecA4AKSTigA6TMBgaQEZEuDZkAfQ68SGi+mTKOcALwA+ThuIxqMYgHUdAC2UxDW8xAEoIxHk4QhIyHCCACWweHggAQnD5fiA):

```ts
import { danger, message, markdown, warn, fail } from "danger"

if (danger.git.modified_files.find(f => f.startsWith(".ts"))) {
  message("Hello!")
}
```

I don't think Swift has this sort of de-composing API which makes this make a bit more sense within the language (I'm generally not a fan of the free floating (hard to find) functions either)

417-72KI:
@orta Sorry for late, and couldn't understand what does `de-composing API` mean.
BTW, I created these changes in this PR as [plugin](https://github.com/417-72KI/DangerSwiftPlugin), so close this PR.

orta:
Cool!

PR #333, opened 2020-04-01
Support cwd parameter

f-meloni:
Many people prefers using the SPM Danger configuration, because is more performing.
Having a Package.swift on your root folder can be annoying, especially now that Xcode (from 11) doesn't put on the recents list an `xcproj` (or `xcworkspace`) when there is a `Package.swift` in the same folder.
At the moment if you move the `Package.swift` in another folder Danger breaks because the path are relatives to the root folder.
Of course the paths can be changed on the Dangerfile in order to have `..` or the full path, but this would not work for example for Plugins where the path is automatically taken from Danger's modified files.
This is why I thought would be nice to offer a `current work directory` option to make you run danger like it was in the root folder of your project even if you have it somewhere else.
I can also think this could be useful in cases like https://github.com/WeTransfer/WeTransfer-iOS-CI/tree/master/Danger-Swift where you have your CI configuration in a separated repo.

PS. this change to work requires the changes on https://github.com/danger/danger-js/pull/1021
@orta can you please release a new danger-js version?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,17 @@
 
 ## Master
 
-- Allow the usage of `~>` to pin a dependency to a major version  [@f-meloni][] - [#326](https://github.com/danger/swift/pull/326)
+## 3.2.2
+
+- Fix danger-swift edit [@lucasmpaim][] - [#332](https://github.com/danger/swift/pull/332)
+
+## 3.2.1
+
+- Fix make install [@lucasmpaim][] - [#331](https://github.com/danger/swift/pull/331)
+
+## 3.2.0
+
+- Update OctoKit to support Xcode 11.4 [@lucasmpaim][] - [#329](https://github.com/danger/swift/pull/329)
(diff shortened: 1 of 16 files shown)


danger-public:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  **Sources/RunnerLib/SPMDanger.swift#L53** - Don't add a space between the method name and the parentheses. (`no_space_in_method_call`)
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against e5096f9134c24599d5e8a3ab41165393ebdbbae8
</p>

orta:
Looks good overall, but it's not a feature unless it's documented ;)

Issue #336, opened 2020-04-09
[Question] Do we need to install the danger-swift when using SPM?


amadeu01:
I have set up the swift package manager file

```swift
// swift-tools-version:5.2
// The swift-tools-version declares the minimum version of Swift required to build this package.

import PackageDescription

let package = Package(
    name: "DangerTarget",
    products: [
        .library(name: "DangerDepsPipefy", type: .dynamic, targets: ["DangerTarget"])
    ],
    dependencies: [
       .package(name: "danger-swift", url: "https://github.com/danger/swift.git", from: "3.3.0")
    ],
    targets: [
        .target(
            name: "DangerTarget",
            dependencies: [.product(name: "danger-swift", package: "danger-swift")],
            path: ".", 
            sources: ["dummy.swift"]
        ),
    ]
)

```

When I try to run the `swift run danger-swift command`

```
👉  swift run danger-swift local
ERROR: Danger JS was not found on the system
Please install it with npm or brew
```

I'm getting an error, which is not clear in the setup whether I have to install the dependencies with homebrew or not. 

From docs: 

> We recommend you install Danger via Swift Package Manager. Though you can use homebrew also. Using Swift PM means you safely lock your versions, making your tooling more reliable.

It appears to me that with swift package manager I would be able to run the danger without having to install via homebrew. Did I get it wrong?

orta:
It's asking for Danger JS - not Danger Swift, which it sounds like you've installed correctly 

> ERROR: Danger JS was not found on the system

I'm pretty sure that's in the getting started guides?

amadeu01:
Actually, you're right. However, it the begging of the guide, it let you think that you would only need the swift package build and it would handle the dependencies (which I thought it would be the js part of the danger as well)

![image](https://user-images.githubusercontent.com/11631092/78934517-732e3780-7a81-11ea-8fef-92af4c8f8820.png)

```
Swift PM
You’ll need to be using Xcode 10 or above. We need to create a package definition, so create a Package.swift in the root of your folder:

// swift-tools-version:4.2
import PackageDescription

let package = Package(
    name: "Eigen",
    dependencies: [
      .package(url: "https://github.com/danger/swift.git", from: "1.0.0")
    ],
    targets: [
        // This is just an arbitrary Swift file in our app, that has
        // no dependencies outside of Foundation, the dependencies section
        // ensures that the library for Danger gets build also.
[...]

PR #337, opened 2020-04-09
Update README

amadeu01:
To avoid confusions related to the dependency of Danger JS  and be sure the user knows that they have to install Danger JS  to be able to run swift-danger with SwiftPM

Related to #336


f-meloni:
merge on green

PR #344, opened 2020-05-02
Fix release package action

f-meloni:


orta:
Seems reasonable to me

PR #367, opened 2020-09-05
Use GitHub Container Registry

f-meloni:


.github/workflows/publish_package.yml
```diff
@@ -16,13 +16,13 @@ jobs:
       run: echo ::set-env name=VERSION::${GITHUB_REF/refs\/tags\//}
     
     - name: Docker Login
-      run: echo $PACKAGES_WRITE_TOKEN | docker login docker.pkg.github.com -u $USERNAME --password-stdin
+      run: echo $PACKAGES_WRITE_TOKEN | docker login ghcr.io -u $USERNAME --password-stdin
       env:
         PACKAGES_WRITE_TOKEN: ${{ secrets.DOCKER_GITHUB_PASSWORD }}
         USERNAME: ${{ secrets.DOCKER_GITHUB_USERNAME }}
```
orta, line 22: I assume these may need to change too

orta (approved):

.github/workflows/publish_package.yml
f-meloni: One of the nice things of the Container Registry is that it needs the same permissions (write and read for packages), then the same GitHub token can be used 💪

PR #377, opened 2020-09-21
make SwiftLintViolation properties public

feighter09:
In our project, we use swiftlint to flag todos as warnings in addition to style warnings. For todos, we have a limit of ~10 before we want to block the build, while we don't want any style / compiler warnings allowed, so I'd like to write a danger rule like so:
```
let violations = SwiftLint.lint(.all(directory: nil))
let todosLimit = 10

let todos = violations.filter { $0.ruleID == "todo_rule" }
let warnings = violations.count - todos.count

if warnings.count > 0 {
    fail("Warning count exceeds \(warningsLimit) warnings")
}
if todos.count > todosLimit {
    fail("todo count exceeds \(todosLimit) todos")
}
```

but currently cannot as `ruleID` is marked `internal`

### DIFF
--- a/Sources/Danger/Plugins/SwiftLint/SwiftLintViolation.swift
+++ b/Sources/Danger/Plugins/SwiftLint/SwiftLintViolation.swift
@@ -4,11 +4,11 @@ public struct SwiftLintViolation: Decodable {
         case error = "Error"
     }
 
-    public var ruleID: String
-    public var reason: String
-    public var line: Int
-    public var severity: Severity
-    public var file: String
+    public internal(set) var ruleID: String
+    public internal(set) var reason: String
+    public internal(set) var line: Int
+    public internal(set) var severity: Severity
+    public internal(set) var file: String
 
     var messageText: String {
         reason + " (`\(ruleID)`)"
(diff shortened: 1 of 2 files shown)


danger-public:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against e8b9cd1903295e4fedae57c2d166f8ce50a6429a
</p>

Sources/Danger/Plugins/SwiftLint/SwiftLintViolation.swift
```diff
@@ -1,14 +1,14 @@
...
-    var line: Int
-    var severity: Severity
-    var file: String
+    public internal(set) var ruleID: String
+    public internal(set) var reason: String
+    public internal(set) var line: Int
+    public internal(set) var severity: Severity
+    public internal(set) var file: String
 
     var messageText: String {
```
417-72KI, line 13: I think this also can be public 🤔

f-meloni: I think this is mainly for the PR message, so given they are automatically posted I think this can stay internal

f-meloni (approved):
Thank you for the PR, I think this is a reasonable change, we are returning the violations so makes sense to make them usable

PR #380, opened 2020-09-26
Release Docker image with SwiftLint

f-meloni:
A lot of people uses the SwiftLint plugin, so I though it would make sense to make an image with SwiftLint in it.
I still want to keep the current image small though, so I'm creating a different image after updating the same Dockerfile.


danger-public:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against 2d1e421a891e926f7b35aa53f806c99be6304876
</p>

Dockerfile
```diff
@@ -16,9 +16,8 @@ RUN apt-get update -q \
     && rm -r /var/lib/apt/lists/*
 
 
-# ARG SWIFT_LINT_VER=0.39.1 # swiftlint
-# RUN git clone -b $SWIFT_LINT_VER --single-branch --depth 1 https://github.com/realm/SwiftLint.git _SwiftLint # swiftlint
-# RUN cd _SwiftLint && git submodule update --init --recursive; make install # swiftlint
+# ARG SWIFT_LINT_VER=0.40.3 # swiftlint
```
vknabel, line 19: I got an error here: `Error response from daemon: Dockerfile parse error line 19: ARG requires exactly one argument`

After removing `# swiftlint` it worked fine. Though this would require the `sed`-command to be changed.
No problem at the next line.

vknabel (commented):
Great feature and nice to get rid of the commits and tags!

Dockerfile
f-meloni: Really interesting 🤔 
The reason why the swiftlint is needed is to tell the script which line uncomment, I will give it a look.
Thank you very much for checking it :)

f-meloni: I've solved this by adding the version to the next line

vknabel (approved):
Looks fine from my perspective 👍

kinwahlai:
hi, this may sound like a stupid question. 
how can I use this docker image with other danger plugins like code coverage or xcodesummary? is it possible?

f-meloni:
Hi @kinwahlai,
It should always possible to use a danger plugin with Danger, also in a docker image.
The easier way I suppose is the SPM configuration, but should work also with Danger's resolver.
You have to be careful though about performances, so you might want to be sure to cache folder containing the downloaded plugins and restore them on the image on the next execution.
Also for the plugins you mentioned you'll need to copy the build artefacts generated by xcodebuild/SPM needed by the plugin in the docker image.

PR #413, opened 2021-02-21
Support package path for swiftlint

f-meloni:


f-meloni:
@orta I think there is something wrong with Danger Public's token
I'm getting

```
Response: {
  "message": "Bad credentials",
  "documentation_url": "https://docs.github.com/rest"
}
```

orta:
Maybe the token got revoked at GitHub level

PR #418, opened 2021-02-28
Test the minimum Swift version supported (5.1.1)

lunij:
This is expected to fail for Linux because of `swiftenv` cannot build version `5.1.1`.


lunij:
Problem will be solved with #419 by switching the Swift version to 5.2.4 that is supported by `swiftenv`.

PR #419, opened 2021-02-28
Switch to Swift 5.2

lunij:
Please see #414 for more information


lunij:
@orta @f-meloni How do you decide about the minimum Swift version? Should it be supported by `swiftenv`? In case of `5.2` it would be `5.2.4`.

f-meloni:
5.2.4 looks good to me as swiftenv version, as long as SPM declares 5.2 it should be fine

f-meloni (approved):
Thank you, I will make a release before merging that will be the latest with swift 5.1 support

PR #424, opened 2021-03-10
Replace Moya/Harvey reference with Moya/Moya

sunshinejr:
Was going through the docs and saw this one mentioning Harvey's config ([*]) and so this updates it to Moya's instead.


f-meloni (approved):
Thank you :)

PR #436, opened 2021-05-02
Edit documentation for GitLab

RobbeRamon:
Edit the example attributes to the correct attributes available (for GitLab).


f-meloni (approved):
Thank you very much!

RobbeRamon:
I have the feeling that the attributes that I found do not match with the DSL. 
These were the attributes I was able to track:

danger-gitlab-attributes

f-meloni (commented):
Looking at https://github.com/danger/swift/blob/master/Sources/Danger/GitLabDSL.swift and https://github.com/danger/swift/blob/master/Sources/Danger/DangerDSL.swift#L22 looks correct

PR #440, opened 2021-05-17
Update README

clayellis-cricut:
I added a few missing details to the README when using the `--cwd` parameter, as well as pointed out that Danger will automatically search for your Dangerfile in a `danger` directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,13 +220,19 @@ danger-swift command --danger-js-path path/to/danger-js
 
 #### Current working directory
 
-Many people prefers using the SPM Danger configuration, because is more performing.
-But having a `Package.swift` on your root folder can be annoying, especially now that Xcode (from 11) doesn't put on the recents list an `xcproj` (or `xcworkspace`) when there is a `Package.swift` in the same folder.
+Many people prefer using the SPM Danger configuration, because is more performant.
+But having a `Package.swift` in your root folder can be annoying, especially now that Xcode (from 11) doesn't put an `xcproj` (or `xcworkspace`) on the recents list when there is a `Package.swift` in the same folder.
 With the `--cwd` parameter you can specify a working directory.
-This allows you to have your `Package.swift` in another directory and still run danger-swift as it was executed from your project root directory.
+This allows you to have your `Package.swift` in another directory and still run danger-swift as it was executed from your project root directory. Since you no longer have a `Package.swift` in your root directory, you'll also have to specify that package path to `swift run`.
 
 ```swift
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -220,13 +220,19 @@ danger-swift command --danger-js-path path/to/danger-js
...
-But having a `Package.swift` on your root folder can be annoying, especially now that Xcode (from 11) doesn't put on the recents list an `xcproj` (or `xcworkspace`) when there is a `Package.swift` in the same folder.
+Many people prefer using the SPM Danger configuration because is more performant.
+But having a `Package.swift` in your root folder can be annoying, especially now that Xcode (from 11) doesn't put an `xcproj` (or `xcworkspace`) on the recents list when there is a `Package.swift` in the same folder.
 With the `--cwd` parameter you can specify a working directory.
-This allows you to have your `Package.swift` in another directory and still run danger-swift as it was executed from your project root directory.
+This allows you to have your `Package.swift` in another directory and still run danger-swift as it was executed from your project root directory. Since you no longer have a `Package.swift` in your root directory, you'll also have to specify that package path to `swift run`.
 
 ```swift
-swift run danger-swift command --cwd path/to/working-directory
+swift run --package-path path/to/working-directory danger-swift command --cwd path/to/working-directory
```
f-meloni, line 229: `--package-path` and `--cwd` do almost the same thing, but in different ways (so you shouldn't need `--cwd` if you have `--package-path`), but `--package-path` doesn't work in all the cases, it has still some corner cases that needs to be handled, and this is why it is not suggested yet, once it fully supported cwd should be deprecated.

The correct command should be `cd path && swift run danger ci --cwd rootFolder`

README.md
```diff
@@ -220,13 +220,19 @@ danger-swift command --danger-js-path path/to/danger-js
...
+```
+
+Keep in mind that Danger will automatically search for your Dangerfile in a `danger` directory placed at the root of your project. So if you put your `Dangerfile.swift` and `Package.swift` in a `danger` directory you can leave off the `--cwd` parameter (though, you still have to specify your package path to `swift run`)
```
f-meloni, line 232: You can use the `--dangerfile` parameter to specify the Dangerfile you want to run, there is no need to use cwd in that case I think, you can use `--package-path` and `--dangerfile`

f-meloni:
Thank you for the PR, just some comments about the --package-path :)

README.md
417-72KI: IMHO, there is no information about `--dangerfile` parameter in any reference or documents, and should be written in the README🤔.

PR #443, opened 2021-05-28
Fix docs scripts

f-meloni:


Dangerfile.swift
```diff
@@ -20,7 +20,9 @@ if !swiftFilesWithCopyright.isEmpty {
     warn("In Danger JS we don't include copyright headers, found them in: \(files)")
 }
 
-SwiftLint.lint(.modifiedAndCreatedFiles(directory: nil), inline: true)
+let filesToLint = danger.git.modifiedFiles + danger.git.createdFiles.filter { !$0.contains("Documentation/") }
```
417-72KI, line 23: It seems that new files in Documentation directory will be ignored, **but modified documentation files are not**.
Is it expected?

417-72KI: If files in `Documentation` will always be ignored, it can be 

```suggestion
let filesToLint = (danger.git.modifiedFiles + danger.git.createdFiles).filter { !$0.contains("Documentation/") }
```

f-meloni: Thank you 🤦🏻 
I wrote the `()` after I already cancelled the danger-swift edit and it was not saved on the original file

PR #446, opened 2021-06-21
Add Xcode 12.5 and Swift 5.4 to CI

f-meloni:


f-meloni:
I've requested access to macOS 11 so that we can have CI there too, I realised we missed that check since when we moved from Travis, and I want to be sure it works well also on the latest version of Xcode.

PR #458, opened 2021-08-14
Add Homebrew path for Apple Silicon based macOS installations

majd:
The Homebrew install path on Apple Silicon is different from the Intel-based Homebrew setup. This PR adds the new path to `Runtime.swift` to fix the error below.

```
ERROR: Could not find a libDanger to link against at any of: [".build/debug", ".build/x86_64-unknown-linux/debug", ".build/release", "/usr/local/lib/danger"]
```

### DIFF
--- a/Sources/RunnerLib/Runtime.swift
+++ b/Sources/RunnerLib/Runtime.swift
@@ -20,7 +20,8 @@ public enum Runtime {
         ".build/debug", // Working in Xcode / CLI
         ".build/x86_64-unknown-linux/debug", // Danger Swift's CI
         ".build/release", // Testing prod
-        "/usr/local/lib/danger" // Homebrew installs lib stuff to here
+        "/usr/local/lib/danger", // Intel Homebrew installs lib stuff to here
+        "/opt/homebrew/lib/danger" // Apple Silicon Homebrew installs lib stuff to here
     ]
 
     /// Finds a path to add at runtime to the compiler, which links


orta (approved):
Looks good to me

majd:
@f-meloni Sorry. I haven't had a chance to get to updating the changelog. Should I still do it or will you take care of that?

To add full Apple Silicon compatibility, in addition to the change made in this PR, the tool itself will need to be compiled universally for both Apple Silicon and Intel.

f-meloni:
@majd No problem at all, I've updated it https://github.com/danger/swift/pull/461/files
For the compilation, brew already uses make install, which installs it using your current machine's Xcode, so I suppose it should make a version that works on your current machine 

https://github.com/danger/homebrew-tap/blob/master/danger-swift.rb#L15
https://github.com/danger/swift/blob/master/Scripts/install.sh

Is there anything missing there?

majd:
I believe one of the libraries is still not being compiled for the current architecture. Here is the log.

```
Starting Danger PR on <repo>
You don't have a DANGER_GITHUB_API_TOKEN set up, this is optional, but TBH, you want to do this
Check out: http://danger.systems/js/guides/the_dangerfile.html#working-on-your-dangerfile
/var/folders/5v/5hst580x74zf9v5vm953v89r0000gn/T/_tmp_dangerfile.swift:1:8: error: module 'Danger' was created for incompatible target arm64-apple-macosx10.10: /opt/homebrew/lib/danger/Danger.swiftmodule
```

I'm happy to take a look if needed. But do you have any ideas? I should probably open a separate issue regarding this.

majd:
@f-meloni I can confirm that the issue I mentioned above has been fixed in #462.

f-meloni:
@majd thank you! I will make a release then!

PR #480, opened 2021-10-25
Add support for Swift 5.5 and Xcode 13

### LINKED ISSUE #475: Latest release unhides DangerDeps product causing failures
This is the error I'm getting:
image

> Swift package target 'Danger' is linked as a static library by 'WeTransferPRLinterTests' and 'Danger', but cannot be built dynamically because there is a package product with the same name.

Using Xcode 13. The plugin is open-source, so you can try it out: 
https://github.com/WeTransfer/WeTransfer-iOS-CI/tree/master/WeTransferPRLinter

This is currently how the Package.swift looks:

```swift
// swift-tools-version:5.5
// The swift-tools-version declares the minimum version of Swift required to build this package.

import PackageDescription

let package = Package(
    name: "WeTransferPRLinter",
    products: [
        .library(
            name: "WeTransferPRLinter",
            targets: ["WeTransferPRLinter"]
        )
    ],
    dependencies: [
        .package(url: "https://github.com/danger/swift", .exact("3.10.2")),
        .package(name: "DangerSwiftCoverage", url: "https://github.com/f-meloni/danger-swift-coverage", from: "1.1.0"),
        .package(name: "DangerXCodeSummary", url: "https://github.com/f-meloni/danger-swift-xcodesummary", from: "1.2.1"),
        .package(name: "Files", url: "https://github.com/JohnSundell/Files", from: "4.1.1")
    ],
    targets: [
        .target(
            name: "WeTransferPRLinter",
            dependencies: [
                .product(name: "Danger", package: "swift"),
                "DangerSwiftCoverage",
                "DangerXCodeSummary",
                "Files"
            ]
        ),
        .testTarget(
            name: "WeTransferPRLinterTests",
            dependencies: [
                "WeTransferPRLinter",
                .product(name: "DangerFixtures", package: "swift")
            ]
        )
    ]
)
```

Commenting out the test target lets Xcode build successfully.

AvdLee:
The `dynamic` type is no longer needed and actually breaks running tests for Danger plugins with Xcode 13.

The error that's fixed with this PR:

> Swift package target 'Danger' is linked as a static library by 'WeTransferPRLinterTests' and 'Danger', but cannot be built dynamically because there is a package product with the same name.

Fixes #475

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@
 
 ## Master
 
+## Next
+- Add support for Swift 5.5 and Xcode 13 [@avdlee][] - [#480](https://github.com/danger/swift/pull/480)
+
 ## 3.12.1
 
 - Comment Danger's internal framework on release [@f-meloni][] - [#477](https://github.com/danger/swift/pull/477)


f-meloni:
I've reproduced the issue here https://github.com/danger/swift/runs/4055493568?check_suite_focus=true and added a CI suite that can reproduce the issue

AvdLee:
@f-meloni is there a middle ground you think that fixes the Xcode 13 bug ánd doesn't break old systems?
As a last resort we could keep using our own fork, but that's obviously not ideal!

f-meloni:
@AvdLee I would love to find something like a parameter that I can pass to swift build to force it to build Danger as dynamic without having to specify it on SPM.
If that doesn't work then the only other solution I can think of is changing the Package.swift from the install script before installing

AvdLee:
> If that doesn't work then the only other solution I can think of is changing the Package.swift from the install script before installing

If it's only failing for that script, it would make total sense. It would be weird to me to not support SPM which is heavily used in Danger environments, so we need to get to a working solution. This sounds like the way to go to me!

PR #485, opened 2021-11-14
Add dynamic only on install

f-meloni:


AvdLee (approved):
🙌  Thanks for taking care of this!

PR #491, opened 2021-12-02
Extract dev dependencies to switch development and release easily

417-72KI:
In this repo, dependencies for development are commented out when release and uncommented soon.
Unfortunately, there has been a fatal issue caused by unhiding Danger Deps (#475).

I think it was a human error, and this PR can avoid it by centralizing management by a simple flag instead of commenting out multiple lines.

When release, we just switch a flag to `false` to make it same condition as previous tags😎

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -78,8 +78,8 @@
         "repositoryURL": "https://github.com/SwiftDocOrg/Markup.git",
         "state": {
           "branch": null,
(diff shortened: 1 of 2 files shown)


f-meloni (approved):
Cool! Will work on a script that changes that automatically during release, so that we are sure is always false when releasing :)

Thank you

PR #495, opened 2022-01-07
Add support for labels on GitHub pull requests

marcomasser:
Danger Swift already supported GitHub labels, but only on issues, not pull requests. This PR adds support for those, too.

Please let me know if it’s OK to just drop this PR here or if there’s anything else that needs to be done on my part.

### DIFF
--- a/Sources/Danger/GitHubDSL.swift
+++ b/Sources/Danger/GitHubDSL.swift
@@ -55,6 +55,7 @@ public extension GitHub {
             case htmlUrl = "html_url"
             case draft
             case links = "_links"
+            case labels
         }
 
         public enum PullRequestState: String, Decodable {
@@ -138,6 +139,9 @@ public extension GitHub {
 
         /// Possible link relations
         public let links: Link
+
+        /// The labels associated with this pull request.
+        public let labels: [Label]
     }
 }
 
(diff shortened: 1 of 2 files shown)


Sources/Danger/GitHubDSL.swift
```diff
@@ -418,19 +422,7 @@ public extension GitHub {
...
-            /// The URL that links to this label.
-            public let url: String
-
-            /// The name of the label.
-            public let name: String
-
-            /// The color associated with this label.
-            public let color: String
-        }
+        public typealias Label = GitHub.Label
```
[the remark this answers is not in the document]
f-meloni: Looks good! Thank you

Sources/Danger/GitHubDSL.swift
```diff
@@ -138,6 +139,9 @@ public extension GitHub {
 
         /// Possible link relations
         public let links: Link
+
+        /// The labels associated with this pull request.
+        public let labels: [Label]
```
f-meloni, line 144: Are the labels actually coming here in the JSON?
Every pull request is also an issue (this is how GitHub handles PRs I think), so `github.issue` should be present for every pull request, so you can get the labels from there I support

marcomasser: Ah, you’re right. I read that every PR is also an issue but somehow haven’t thought about using `github.issue` in my Dangerfile. Thanks for the hint and your quick answers, I’ll close this PR in this case. Wouldn’t be a good idea to deviate from the GitHub API here, I think.

marcomasser: But to answer your question: Yes, the labels are actually coming in the JSON. Even the JSON for the unit test that I updated in this PR has one.

PR #503, opened 2022-01-29
Danger doesn't work with Mint, removing it

f-meloni:


colejd:
Hello! @f-meloni, when you get a chance, could you please explain why you removed this? Is there a technical limitation with Mint that prevents Danger from working with it? I know SPM presents some limitations sometimes, and all you can do is hope Apple fixes it someday - otherwise, I'd like to get a PR going for you to fix the issue if it's possible.

PR #505, opened 2022-02-04
Notify a new release on running `danger-swift`

417-72KI:
Some tools (e.g. `carthage`, `gh`, etc...) notify when new version is released.
I added a version checker to fetch a latest release in GitHub, and notify if running version is older than latest.

### DIFF
--- a/Sources/RunnerLib/VersionChecker.swift
+++ b/Sources/RunnerLib/VersionChecker.swift
@@ -1,0 +1,50 @@
+import DangerShellExecutor
+import Logger
+import Version
+import Foundation
+
+public enum VersionChecker {
+    public static func checkForUpdate(current currentVersionString: String) {
+        if ProcessInfo.processInfo.environment["DEBUG"] != nil {
+            return
+        }
+        guard let latestVersionString = fetchLatestVersion() else { return }
+        guard let latestVersion = Version(latestVersionString) else {
+            logger.debug("Invalid latestVersionString: (\(latestVersionString)")
+            return
+        }
+        guard let currentVersion = Version(currentVersionString) else {
+            logger.debug("Invalid currentVersionString: (\(currentVersionString)")
+            return
+        }
+        if currentVersion < latestVersion {
+            logger.logInfo("\nℹ️  A new release of danger-swift is available: \(currentVersion) -> \(latestVersion)")
+        }
+    }
+
+    static func fetchLatestVersion() -> String? {
+        let decoder = JSONDecoder()
+        decoder.keyDecodingStrategy = .convertFromSnakeCase
+        do {
+            let latest = try ShellExecutor().execute("curl",
+                                                     arguments: [
+                                                        "-s",
+                                                        "https://api.github.com/repos/danger/swift/releases/latest"
+                                                     ])
+                .data(using: .utf8)
+                .flatMap { try decoder.decode(Release.self, from: $0) }
+            return latest?.tagName
+        } catch {
+            logger.debug(error)
+            return nil
+        }
+    }
+}
+
+extension VersionChecker {
+    static let logger = Logger()
+}
+
+struct Release: Codable {
+    var tagName: String
+}
--- a/Sources/RunnerLib/VersionChecker.swift
+++ b/Sources/RunnerLib/VersionChecker.swift
@@ -5,7 +5,7 @@ import Foundation
 
 public enum VersionChecker {
     public static func checkForUpdate(current currentVersionString: String) {
-        if ProcessInfo.processInfo.environment["DEBUG"] != nil {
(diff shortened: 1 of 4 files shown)


Sources/RunnerLib/VersionChecker.swift
```diff
@@ -0,0 +1,50 @@
...
+        guard let currentVersion = Version(currentVersionString) else {
+            logger.debug("Invalid currentVersionString: (\(currentVersionString)")
+            return
+        }
+        if currentVersion < latestVersion {
+            logger.logInfo("\nℹ️  A new release of danger-swift is available: \(currentVersion) -> \(latestVersion)")
+        }
+    }
+
+    static func fetchLatestVersion() -> String? {
```
f-meloni, line 25: I think this can be private

Sources/RunnerLib/VersionChecker.swift
```diff
@@ -0,0 +1,50 @@
...
+            return latest?.tagName
+        } catch {
+            logger.debug(error)
+            return nil
+        }
+    }
+}
+
+extension VersionChecker {
+    static let logger = Logger()
```
f-meloni, line 45: This Logger won't have the right set up, which is made based on the parameters passed to Danger Swift, not a big deal, but would be nice to get it from the runner if possible (is created at line 25)

Sources/RunnerLib/VersionChecker.swift
```diff
@@ -0,0 +1,50 @@
+import DangerShellExecutor
+import Logger
+import Version
+import Foundation
+
+public enum VersionChecker {
+    public static func checkForUpdate(current currentVersionString: String) {
+        if ProcessInfo.processInfo.environment["DEBUG"] != nil {
```
f-meloni, line 8:
```suggestion
        guard ProcessInfo.processInfo.environment["DEBUG"] == nil else {
```

Nitpick

f-meloni: I would like if possible another env variable which if set can opt out this check, so folks can disable if they want

Sources/RunnerLib/VersionChecker.swift
```diff
@@ -0,0 +1,50 @@
+import DangerShellExecutor
+import Logger
+import Version
+import Foundation
+
+public enum VersionChecker {
+    public static func checkForUpdate(current currentVersionString: String) {
```
f-meloni, line 7: Can inject ShellExecutor and Logger (which anyway can be needed), so that we can also have some tests for this logic?

f-meloni (commented):
Thank you very much for this PR :) Great idea!
Just few minor comments:

Sources/RunnerLib/VersionChecker.swift
417-72KI: Fixed `VersionChecker` enum to class, which make it injectable 😄.

Sources/RunnerLib/VersionChecker.swift
417-72KI: made private!

Sources/RunnerLib/VersionChecker.swift
417-72KI: https://github.com/danger/swift/pull/505#discussion_r800280200 may also resove it 😁

Sources/RunnerLib/VersionChecker.swift
```diff
@@ -0,0 +1,58 @@
...
+import Foundation
+
+public class VersionChecker {
+    private let shellExecutor: ShellExecutor
+    private let logger: Logger
+
+    public init(
+        shellExecutor: ShellExecutor = .init(),
+        logger: Logger
+    ) {
```
f-meloni, line 13: Given this is now injectable, can we also add some tests for this please?

f-meloni (commented):
If you merge master CI should pass.

Sources/RunnerLib/VersionChecker.swift
```diff
@@ -0,0 +1,58 @@
+import DangerShellExecutor
+import Logger
+import Version
+import Foundation
+
+public class VersionChecker {
```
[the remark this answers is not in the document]
417-72KI: fixed!

417-72KI:
@f-meloni 
> If you merge master CI should pass.

There is a problem to build SwiftLint on Linux (5.5.2), which makes CI fail.

Latest version of SwiftLint may be resolve it and I made another PR.
https://github.com/danger/swift/pull/509

Sources/RunnerLib/VersionChecker.swift
417-72KI: done!

f-meloni (approved):
🚀

PR #507, opened 2022-02-12
Update LICENSE

f-meloni:


orta:
If you change it to '-present' then you never had to update it again :D

PR #509, opened 2022-02-19
Bump dependencies to avoid SwiftLint build-error in Swift 5.5.2

417-72KI:
There is some problem that SwiftLint build fails on Swift 5.5.2.
https://github.com/danger/swift/runs/5258863489?check_suite_focus=true#step:10:69

I tried to bump it and found version conflicts on some dependencies, so I make them all to latest.

Additionally, [swift-doc](https://github.com/SwiftDocOrg/swift-doc)is archived and not maintained any more, which causes version conflict too.
It seems to be necessary to run `make docs`, so I comment-out temporary instead of removing.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -18,7 +18,6 @@ let devDependencies: [Package.Dependency] = isDevelop
         .package(url: "https://github.com/Realm/SwiftLint", from: "0.46.2"),
         .package(name: "SnapshotTesting", url: "https://github.com/pointfreeco/swift-snapshot-testing.git", from: "1.7.1"),
         .package(url: "https://github.com/shibapm/Rocket", from: "1.2.1"),
-        // .package(url: "https://github.com/SwiftDocOrg/swift-doc", .branch("1.0.0-rc.1")),
     ] : []
 let devTargets: [Target] = isDevelop
     ? [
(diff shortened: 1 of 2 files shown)


Package.swift
```diff
@@ -13,12 +13,12 @@ let devProducts: [Product] = isDevelop
...
-        .package(url: "https://github.com/nicklockwood/SwiftFormat", from: "0.35.8"),
-        .package(url: "https://github.com/Realm/SwiftLint", from: "0.38.0"),
+        .package(url: "https://github.com/shibapm/Komondor", from: "1.1.3"),
+        .package(url: "https://github.com/nicklockwood/SwiftFormat", from: "0.49.4"),
+        .package(url: "https://github.com/Realm/SwiftLint", from: "0.46.2"),
         .package(name: "SnapshotTesting", url: "https://github.com/pointfreeco/swift-snapshot-testing.git", from: "1.7.1"),
-        .package(url: "https://github.com/shibapm/Rocket", from: "0.4.0"),
-        .package(url: "https://github.com/SwiftDocOrg/swift-doc", .branch("1.0.0-rc.1")),
+        .package(url: "https://github.com/shibapm/Rocket", from: "1.2.1"),
+        // .package(url: "https://github.com/SwiftDocOrg/swift-doc", .branch("1.0.0-rc.1")),
```
f-meloni, line 21: Given this is been archived, I think we can remove it completely, will work on make docs separately

f-meloni (approved):

PR #517, opened 2022-05-19
bump Xcode requirement in Homebrew to 12

417-72KI:
Since latest version of Danger-Swift requires Swift 5.3, it should also require Xcode 12.0

ref: https://developer.apple.com/jp/support/xcode/


f-meloni (approved):
Thank you!

Issue #464, opened 2021-09-20
Danger was created for incompatible target arm64-apple-macosx10.10


tibo:
Hello,

I'm trying to move the Danger step of our CI to a Mac Mini M1 for efficiency (pulling the docker image takes too much time for a short CI) and it doesn't work as expected:
```
$ danger-swift pr https://github.com/danger/swift/pull/146
Starting Danger PR on danger/swift#146
You don't have a DANGER_GITHUB_API_TOKEN set up, this is optional, but TBH, you want to do this
Check out: http://danger.systems/js/guides/the_dangerfile.html#working-on-your-dangerfile
/var/folders/sp/0r2dd4g549s9w7n8mhwlb9q00000gn/T/_tmp_dangerfile.swift:1:8: error: module 'Danger' was created for incompatible target arm64-apple-macosx10.10: /opt/homebrew/lib/danger/Danger.swiftmodule
import Danger
       ^

ERROR: Dangerfile eval failed at Dangerfile.swift
ERROR: Could not get the results JSON file at /var/folders/sp/0r2dd4g549s9w7n8mhwlb9q00000gn/T/danger-response.json

Danger: ⅹ Failing the build, there is 1 fail.
## Failures
danger-swift` failed.
## Markdowns
### Log


sh
/var/folders/sp/0r2dd4g549s9w7n8mhwlb9q00000gn/T/_tmp_dangerfile.swift:1:8: error: module 'Danger' was created for incompatible target arm64-apple-macosx10.10: /opt/homebrew/lib/danger/Danger.swiftmodule
import Danger
       ^
ERROR: Dangerfile eval failed at Dangerfile.swift
ERROR: Could not get the results JSON file at /var/folders/sp/0r2dd4g549s9w7n8mhwlb9q00000gn/T/danger-response.json
```

I've got this issue on the Mac Mini but it's running without any issue on my Macbook (also M1) and I cannot find why.
I already tried to re-install danger-js and danger-swift. I've cleaned the homebrew install and rebuild it. still the same.

Any idea?
Thanks

rseki-sonix:
I got the same error. I changed the target from iPhone to Mac and the build was successful.

![any-ios-device](https://user-images.githubusercontent.com/71474084/134099801-04d6a516-73ae-409f-aa8b-927cad294e61.png)
![mac](https://user-images.githubusercontent.com/71474084/134099818-d296954e-d38f-455a-8d6e-39f00f99e9c3.png)

tibo:
[...]

PR #518, opened 2022-05-19
Add platform requirements on `danger-swift edit`

417-72KI:
When running `danger-swift edit`, Package.swift does not specify macOS version and generated xcodeproj's deployment target will be 10.10.

This PR makes it possible to set Deployment Target based on running machine.

image

I don't know why, but it resolves weirdly errors on `import Danger` (e.g. https://github.com/danger/swift/issues/464#issuecomment-997418557)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,9 @@
 
 ## Master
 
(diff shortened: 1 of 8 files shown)


f-meloni (approved):
Thank you :) Can you please also update the CHANGELOG?

PR #521, opened 2022-05-27
Patch-fix for `error: incompatible target arm64-apple-macosx10.10` on M1 mac

417-72KI:
Fixes: #464 

Note: This is not a fundamental solution, and should be reverted when compiler's bug fixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 ## Master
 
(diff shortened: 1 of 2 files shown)


f-meloni (commented):
Can you please explain why this works? I would expect arm64 to be the right arch given it checks that in the if before.
Can you also please update the CHANGELOG? :)

417-72KI:
Although it has not been confirmed, I think the original issue is caused by some bugs on complier or linker in arm64.
This patch makes it to build and run in x86_64 arch (it means to run on Rosetta 2, and it behaves as if it is running on Intel-mac).

It will be unnecessary if there is other way to build Dangerfile.swift script expectedly or my prediction is wrong.

PR #530, opened 2022-08-02
Suffix temporary directory with unique path

### LINKED ISSUE #529: Danger causing unexpected pipeline failures with concurrency enabled
We run iOS CI pipelines using `gitlab-runner` hosted on MacStudium Mac Minis. When two pipelines run at a similar time we get random failures resulting from our Danger step. Danger is setup as a nested Swift Package, which our CI server builds at the start of each pipeline run.

When the runner clones the repo, it clones to a folder using the runner concurrency identifier. For example, here are two possible folders:

```
/Users/[username]/builds/[runner-id]/0/[project-name]/.git/
/Users/[username]/builds/[runner-id]/1/[project-name]/.git/
```

For the most part this works great, however we are getting failures related to the temporary directory Danger uses. Looking at two failed pipelines, I can see while the clone paths are different, the danger temporary path is the same:

```
$ cd Danger && arch -x86_64 swift run danger-swift ci -i post_build -f --dangerfile DangerfilePostBuild.swift --cwd ../ && cd -
[1/1] Planning build
[0/0] Build complete!
<unknown>:0: error: error opening input file '/var/folders/rm/_fdw_5ys44xb_2gswdqh8c4r0000gq/T/_tmp_dangerfile.swift' (No such file or directory)
ERROR: Dangerfile eval failed at DangerfilePostBuild.swift
Failing the build, there is 1 fail.
```

```
$ cd Danger && arch -x86_64 swift run danger-swift ci -i post_build -f --dangerfile DangerfilePostBuild.swift --cwd ../ && cd -
[1/1] Planning build
[0/0] Build complete!
<unknown>:0: error: error opening input file '/var/folders/rm/_fdw_5ys44xb_2gswdqh8c4r0000gq/T/_tmp_dangerfile.swift' (No such file or directory)
ERROR: Dangerfile eval failed at DangerfilePostBuild.swift
Failing the build, there is 1 fail.
```

Is there a way for us to define the temporary directory to use with Danger? That way we could just nest a `.tmp` folder within the clone directory.

squarefrog:
Fixes #529

Reusing the temporary directory can cause issues with concurrent instances of Danger. By appending a unique identifier we can prevent race conditions when running Danger from concurrent processes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 ## Master
 
 - Drop Swift 5.3 support [@417-72KI][] - [#524](https://github.com/danger/swift/pull/524)
+- Refine runner temporary path to better support concurrent pipelines [@squarefrog][] - [#530](https://github.com/danger/swift/pull/530)
 
 ## 3.13.0
 
--- a/Sources/Runner/Commands/Runner.swift
+++ b/Sources/Runner/Commands/Runner.swift
@@ -9,8 +9,13 @@ func runDanger(logger: Logger) throws {
     // Pull in the JSON from Danger JS
     let standardInput = FileHandle.standardInput
     let fileManager = FileManager.default
-    let tmpPath = NSTemporaryDirectory()
+    let tmpPath = NSTemporaryDirectory() + "danger/\(UUID().uuidString)/"
     let dangerResponsePath = tmpPath + "danger-response.json"
+
+    try fileManager.createDirectory(
+        at: URL(fileURLWithPath: tmpPath),
+        withIntermediateDirectories: true
+    )
 
     // Pull in the JSON from Danger JS
     guard let dangerDSLURL = String(data: standardInput.readDataToEndOfFile(), encoding: .utf8) else {
--- a/Sources/Runner/Commands/Runner.swift
+++ b/Sources/Runner/Commands/Runner.swift
@@ -187,7 +187,7 @@ func runDanger(logger: Logger) throws {
 
     // Clean up after ourselves
     try? fileManager.removeItem(atPath: dslJSONPath)
-    try? fileManager.removeItem(atPath: tmpPath)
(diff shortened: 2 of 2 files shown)


f-meloni (approved):
Thank you! Can you please update the CHANGELOG file as well? :)

squarefrog:
Of course! Looks like I've broken all of the pipelines though :(

squarefrog:
All set for review now @f-meloni, sorry for the CI spam!

When the runner is run, it creates a new temporary directory with a unique ID, for example:

```
/var/folders/86/1vtftp_j2rn_wdcq6232q_z80000gq/T/danger/4FE1BABB-D918-4FB6-9CCF-D27B45034149/
```

The temporary files are then created within this folder. Upon completion (either success or failure), then the unique folder is cleaned up leaving only the parent folder:

```
/var/folders/86/1vtftp_j2rn_wdcq6232q_z80000gq/T/danger/
```

One thing I'm unsure of though, is the `dslJSONPath`. I'm unsure where this is as it's passed from Danger JS, so I've left the cleanup for that as-is.

squarefrog:
OK I don't think is quite there yet, I feel I'm missing some context. If we just delete the files individually, CI passes. If we remove the parent folder instead, half the CI runners fail. Perhaps I'm clearing these files up too soon?

squarefrog:
Do I need to do anything else to get this merged in?

f-meloni:
No, I thought I've merged it already, sorry

squarefrog:
Thanks so much!

squarefrog:
Bit of a long shot, but do you know how we can use this immediately on our pipeline? We tried pointing our Package.swift at the commit we need:

```
.package(name: "danger-swift", url: "https://github.com/danger/swift.git", .revision("263436d79c51bf32e21b960f387bac4ef2d05c34")
```

This failed though, as we also use Danger Xcode Summary, which relies on tagged versions of danger swift.

PR #543, opened 2022-10-25
Make Swift versions and supported Danger-Swift versions list as table matrix instead of bullet points

417-72KI:


f-meloni (approved):
Nice!

PR #572, opened 2023-02-14
GitLab Add deactivated user status

antigp:
I added "deactivated" status to user. Because GitLab can return this status in reviewers array. 
Otherwise it fail to decode all dsl json.


f-meloni (approved):
Thank you :)

f-meloni:
Can you please also update the changelog?

antigp:
Done

Issue #476, opened 2021-10-13
[Feature Request] Include a pre-built executable in each release and use in the homebrew tap


bobergj:
### Summary
Include a pre-built executable in each release. Use that pre-built artifact in the homebrew tap. 

### Why
* Fetching the dependencies and building danger-swift can take a long time, even with the swiftpm approach, especially on CI such as Xcode Cloud where cache is per workflow/branch
* The homebrew tap builds from source https://github.com/danger/swift/blob/master/Scripts/create_homebrew_tap.sh
* The danger-js homebrew tap already serves a pre-built binary https://github.com/danger/homebrew-tap/blob/master/danger-js.rb

### How
 
Building a universal binary for Intel and Apple Silicon macs should be a matter of:
```
swift build --configuration release --arch arm64 --arch x86_64
```

Inspecting the resulting executable:
```
file .build/apple/Products/Release/danger-swift
.build/apple/Products/Release/danger-swift: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64:Mach-O 64-bit executable arm64]
.build/apple/Products/Release/danger-swift (for architecture x86_64):	Mach-O 64-bit executable x86_64
.build/apple/Products/Release/danger-swift (for architecture arm64):	Mach-O 64-bit executable arm64
```

```
du -h .build/apple/Products/Release/danger-swift
1.4M	.build/apple/Products/Release/danger-swift
```

Since the third-party dependencies are statically linked, the executable is standalone:
```
otool -L .build/apple/Products/Release/danger-swift
.build/apple/Products/Release/danger-swift:
	/System/Library/Frameworks/Foundation.framework/Versions/C/Foundation (compatibility version 300.0.0, current version 1775.118.101)
	/usr/lib/libobjc.A.dylib (compatibility version 1.0.0, current version 228.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.100.5)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1775.118.101)
	@rpath/libswiftCore.dylib (compatibility version 1.0.0, current version 1205.0.24)
	@rpath/libswiftCoreFoundation.dylib (compatibility version 1.0.0, current version 1.6.0, weak)
	@rpath/libswiftCoreGraphics.dylib (compatibility version 1.0.0, current version 2.0.0, weak)
[...]

PR #579, opened 2023-03-20
Create example approach to Danger framework with module stability

DavidBrunow:
This pull request shows an example implementation of changes that could be made to enable module stability for the Danger dependency used by Dangerfile.swift. It also builds a universal binary for danger-swift. 

All credit for the idea goes to @bobergj, I saw this comment, needed to implement it for a fork used internally at work, and thought I'd share it in case it is useful: https://github.com/danger/swift/issues/476#issuecomment-984248497

Drawbacks to this approach in general (as mentioned in that issue):
* Needs a project file which could get out of sync with the Package.swift
* `@_implementationOnly` is underscored and is going to change in the future

This specific implementation also has this very large problem:
* It does not solve for the problem of the OctoKit API being exposed to consumers.

I've added a script that shows how danger-swift and the Danger.framework can be built.

### DIFF
--- a/Projects/Danger/Danger.xcodeproj/project.pbxproj
+++ b/Projects/Danger/Danger.xcodeproj/project.pbxproj
@@ -1,0 +1,663 @@
+// !$*UTF8*$!
+{
+	archiveVersion = 1;
+	classes = {
+	};
+	objectVersion = 56;
+	objects = {
+
+/* Begin PBXBuildFile section */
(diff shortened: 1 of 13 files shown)


rarias84:
how I can use this approach?

DavidBrunow:
> how I can use this approach?

If you are using GitHub then you cannot use the current state of this because more work would need to be done to wrap the OctoKit dependency. 

If you are using a different tool, like BitBucket or GitLab, you can run the `build_runner_and_framework.sh` script which will create a `danger_swift` binary and a `Danger.framework` framework. I've only experimented with those being in the same directory when running Danger Swift but theoretically you could place both somewhere in your path and things would "just work".

PR #588, opened 2023-07-02
Use Version struct for danger swift version

f-meloni:


417-72KI (approved):
LGTM, except lint warnings

PR #595, opened 2023-09-21
Update BitBucketServerDSL.swift

dromerobarria:
Update task value to optional to solve the parser in Bitbucket.


f-meloni (approved):
Thank you!

PR #610, opened 2024-06-12
Expose markdownAction in SwiftLint lint function

Nikoloutsos:
**Problem:**
When using the `SwiftLint.lint()` I do not have an option to be able to have my own customized output.

**Solution:**
Expose the markdownAction so that the developer is able to disable the default output and use his own.


f-meloni (approved):
Thanks :)

PR #614, opened 2024-07-12
Drain stdout pipe while command is running

jflan-dd:
One of the commands we called as part of a Danger plugin ended up generating more output that the `Pipe`'s buffer could hold. This resulted in Danger locking up and blocking CI.

This change moves reading from stdin to before the task has exited so that the `Pipe` does not get too full.

You can test this change will the following code:
```swift
let shell = ShellExecutor()
let result = try shell.spawn("hexdump /dev/urandom | head -c 500000", arguments: [])
```

I thought about adding a test, but if the test were to ever go wrong it would lock up CI instead of failing


jflan-dd:
@f-meloni Could you review this when you have a chance?

Sources/DangerShellExecutor/ShellExecutor.swift
```diff
@@ -61,9 +61,11 @@ public struct ShellExecutor: ShellExecuting {
         let pipe = Pipe()
         task.standardOutput = pipe
         task.launch()
-        task.waitUntilExit()
 
         let data = pipe.fileHandleForReading.readDataToEndOfFile()
+
+        task.waitUntilExit()
```
f-meloni, line 67: Would you still receive all the text output coming from the executed shell command after this change?
Given we are reading before the task ends

jflan-dd: We're calling `readDataToEndOfFile()` so we'll get the entire output of the command from the pipe before continuing to `task.waitUntilExit()`, which I assume will basically immediately complete at that point.

jflan-dd: @f-meloni does that seem correct?

f-meloni: Let's give it a chance :)

f-meloni (approved):
Would you mind please updating the Changelog before merging? Thank you

PR #616, opened 2024-08-28
Fix pattern for detecting SwiftLint in package

417-72KI:
Danger-Swift uses SwiftLint in the package when it contains, which takes too much times to fetch and build.
There is a repo which only contains plugins (https://github.com/SimplyDanny/SwiftLintPlugins), `swift run swiftlint` is fail when the package conatins it instead of SwiftLint directly.
However, current spec causes failure ( try to run `swift run swiftlint` and fail ) because 
```
.package(url: "https://github.com/SimplyDanny/SwiftLintPlugins", from: "<version>"),
```
matches to 
```
/\.package\(.*SwiftLint.*/
```

This PR fixes a regex to avoid false positive of `SwiftLintPlugins`.

This is a test case
image


f-meloni (approved):
Thank you!

PR #618, opened 2024-08-31
Update dependencies

417-72KI:


417-72KI:
Already updated in #620

PR #631, opened 2024-11-11
Merge duplicated imports in finally-generated Dangerfile

417-72KI:
Dividing files using `// fileImport:` is a very useful feature on Danger-Swift.

However, importing module in extracted file may causes an error.

### Dangerfile.swift
```swift
import Danger

let danger = Danger()

do {
  // fileImport: DangerExtensions/SomeFeature.swift
  try doSomething(danger, file: "path/to/file")
} catch {
  logger.error(error)
  danger.fail(error.localizedDescription)
}
```

### DangerExtensions/SomeFeature.swift
```swift
import Danger // - error: declaration is only valid at file scope

func doSomething(_ danger: DangerDSL, filePath: String) throws {
  // ...
}
```

This PR makes it to merge duplicated module imports and declare at the top on generated `_tmp_dangerfile.swift`.

### DIFF
--- a/DangerfileExtensions/ChangelogCheck.swift
+++ b/DangerfileExtensions/ChangelogCheck.swift
@@ -1,3 +1,5 @@
+import Danger
+
 func checkChangelog() {
     let allSourceFiles = danger.git.modifiedFiles + danger.git.createdFiles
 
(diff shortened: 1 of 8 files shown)


f-meloni (approved):
Sorry for taking so long to review this!

PR #643, opened 2025-02-24
Updated tutorial documentation to add notes when using swiftlint

stotic-dev:
I discovered the following tricky issue when using SwiftLint, so I added a note about it as a supplement to the tutorial documentation.

If the following setting is included in the SwiftLint configuration file, the JSON output from SwiftLint will contain not only the static analysis results but also messages like "Your version of SwiftLint is up to date." This prevents Danger from correctly parsing the output, causing its execution to fail.

```yml
check_for_updates: true  
```


f-meloni (approved):
Thank you!

PR #646, opened 2025-03-16
Fix detecting `MAJOR_VERSION` in linux

417-72KI:
When `swift --version` returns like `6.0.3` installation script fails to get `MAJOR_VERSION` and modules are not copied into `LIB_INSTALL_PATH`.

### DIFF
--- a/Scripts/install.sh
+++ b/Scripts/install.sh
@@ -20,7 +20,7 @@ if [[ "$OSTYPE" == "darwin"* ]]; then
 else
     BUILD_FOLDER=".build/debug"
     swift build --disable-sandbox
-    MAJOR_VERSION=$(swift --version | awk '{for(i=1;i<=NF;i++){if($i ~ /^[0-9]+\.[0-9]+\.[0-9]?$/){print $i; break}}}' | cut -d '.' -f 1)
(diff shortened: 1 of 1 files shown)


Scripts/install.sh
```diff
@@ -20,7 +20,7 @@ if [[ "$OSTYPE" == "darwin"* ]]; then
 else
     BUILD_FOLDER=".build/debug"
     swift build --disable-sandbox
-    MAJOR_VERSION=$(swift --version | awk '{for(i=1;i<=NF;i++){if($i ~ /^[0-9]+\.[0-9]+$/){print $i; break}}}' | cut -d '.' -f 1)
+    MAJOR_VERSION=$(swift --version | awk '{for(i=1;i<=NF;i++){if($i ~ /^[0-9]+\.[0-9]+\.[0-9]?$/){print $i; break}}}' | cut -d '.' -f 1)
```
f-meloni, line 23: I'm not a regex expert, but shouldn't this be (\.[0-9]+)? To have all the group optional and support more than one digit?

417-72KI: Certainly, thank you.

417-72KI: fixed!

f-meloni (approved):

PR #647, opened 2025-04-08
Fixed issue with string force unwrap

Davarg:
There may be situation when file doesn't contain EOF symbol. So string force unwrapping will cause crash.

### DIFF
--- a/Sources/DangerShellExecutor/ShellExecutor.swift
+++ b/Sources/DangerShellExecutor/ShellExecutor.swift
@@ -77,8 +77,8 @@ public struct ShellExecutor: ShellExecuting {
 
             task.waitUntilExit()
 
-            let result = String(data: data, encoding: .utf8) ?? ""
-            return result.trimmingCharacters(in: .whitespacesAndNewlines)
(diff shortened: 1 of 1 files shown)


Sources/DangerShellExecutor/ShellExecutor.swift
```diff
@@ -77,7 +77,8 @@ public struct ShellExecutor: ShellExecuting {
 
             task.waitUntilExit()
 
-            return String(data: data, encoding: .utf8)!.trimmingCharacters(in: .whitespacesAndNewlines)
+            let result = String(data: data, encoding: .utf8) ?? ""
+            return result.trimmingCharacters(in: .whitespacesAndNewlines)
```
f-meloni, line 81: Nit

```suggestion
            let result = String(data: data, encoding: .utf8).map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }  ?? ""
            return result
```

So you do the trimming only if not nil

f-meloni (approved):

Sources/DangerShellExecutor/ShellExecutor.swift
Davarg: fixed

PR #659, opened 2025-11-13
Fix enum case value for mannequin in UserType

tahirmt:
The user type returned from the GitHub API has a different case than what is expected here. 

Screenshot 2025-11-13 at 12 15 24 PM


f-meloni (approved):

tahirmt:
@f-meloni thanks for merging. Any chance of getting a version published with this change?

PR #660, opened 2026-06-04
feat: publish prebuilt universal macOS binary on release

nenadvulic:
## Problem

The existing `danger/tap/danger-swift` Homebrew formula works locally, but has two issues on CI runners:

1. **`brew tap danger/tap` requires a git clone via HTTPS** — this fails silently on environments with restricted network access (e.g. Xcode Cloud), leaving `danger-swift` unavailable with no clear error message.

2. **The formula compiles from source** (`make install`) — this takes ~10 minutes and requires Xcode as a build dependency on every clean CI runner.

## Solution

This PR adds a `release_macos_binary.yml` workflow that runs on every tag and:
1. Builds `danger-swift` for both `arm64` and `x86_64`
2. Merges them into a **universal binary** using `lipo`
3. Uploads `danger-swift-macos-universal.tar.gz` + its SHA256 to the GitHub Release

It also updates `Scripts/create_homebrew_tap.sh` to generate a formula that installs the prebuilt binary directly — removing the Xcode build dependency and reducing install time from ~10 minutes to a few seconds.

## Result

```bash
brew install danger/tap/danger-swift  # now takes ~5 seconds
```

The universal binary works natively on Apple Silicon and under Rosetta (x86_64) without needing `arch -arm64` workarounds.


.github/workflows/release_macos_binary.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Package binary
+        run: |
+          VERSION="${GITHUB_REF_NAME}"
+          TARBALL="danger-swift-macos-universal.tar.gz"
+          tar -czf "$TARBALL" -C .build/release danger-swift
+          shasum -a 256 "$TARBALL" > "$TARBALL.sha256"
+          echo "TARBALL=$TARBALL" >> "$GITHUB_ENV"
+          echo "VERSION=$VERSION" >> "$GITHUB_ENV"
+
+      - name: Upload to GitHub Release
```
f-meloni: I think we already have a job to create a release today, should we add this to that one instead?

f-meloni (commented):
Thanks for the PR! Just a comment

.github/workflows/release_macos_binary.yml
f-meloni: This one https://github.com/danger/swift/blob/master/.github/workflows/create_release.yml

nenadvulic: @f-meloni  Good call,  i folded the macOS build into create_release.yml as a release-macos job gated on needs: build, so the binary uploads after the release is created. 

This also removes a race: both workflows triggered on the same tag and each created the release (gh release create vs action-gh-release). Standalone workflow deleted.

f-meloni (approved):

